PR #70, opened 2022-10-15
Feat/apk download

### LINKED ISSUE #69: [Android] Add programatic way to download latest apk

chrismclarke:
## TODO

- [ ] Decide best place to host/store APKs
- [ ] Add cloud function to retrieve from picsa.app/apk or picsa.app/dl (or link in app?)
- [ ] Automate picking version code
- [ ] Integrate generate with publish
- [ ] Publish updates to play store

## Description
When publishing updates there is usually a delay between submitting updates to the play store and them being approved, usually this is just 1-2 days but sometimes can be as long as a week. This means a fallback is usually required to share the latest app as an apk. 

There are 2 challenges with this approach:
1) Sharing is typically this is done via whatsapp groups, however not all users have whatsapp on their work devices and so then have a second step of transferring via bluetooth.

2) Generating the apk is a slightly involved process, as it requires developer access to the google play account and navigating various submenus to locate the universal apk. 

This PR attempts to address these issues by automating the process of generating apks from the play store, and making available for direct download from the website https://picsa.app/apk

Closes #69

## Discussion

## Preview

https://picsa.app

## Screenshots / Videos


chrismclarke:
This could be revisited in the future, but for now assume manually downloading from play console fine. May decided to revisit in the future if integrating with dashboard

PR #92, opened 2023-01-25
Nextra Themed Docs

FaithDaka:
## Description

- New _pages_ folder which includes the documentation files transferred from the **documentation** folder
- New _public_ folder which includes public assets
- New packages for next and nextra docs
- Configuration files to setup next and nextra

## Discussion

- These docs were created from copy and pasting from existing docs. Will need updates.

## Preview

https://picsa.app

## Screenshots / Videos
image

image


FaithDaka:
Ready for review @chrismclarke

chrismclarke (changes requested):
Thanks @FaithDaka 
I've tested locally and it runs smoothly, I also appreciate the additional sections you've added for each of the applications.

Looking at the current folder structures, I think having all the nextJS code at the top-level of this repo will probably make things more confusing in the long run. So instead I've gone ahead and created a new repo for the documentation, which should also make it easier to add custom deployment scripts and keep the list of dependencies in the main package.json lower.

I've invited you as a collaborator, would you be able to repeat these steps on the new repo and open as a PR there instead?

https://github.com/e-picsa/docs

FaithDaka:
@chrismclarke Thanks for the separation of concerns.
I should be able to transfer the changes to the _docs_ repository.
I'd seen the confusion as well since the NextJS architecture requires a **pages** root folder for navigation, so this resolution will be better.

PR #103, opened 2023-02-13
adding an option functionality

khalifan-kfan:
## Description

Implement the initial idea of the options tool, basically, a user enters the option in terms of seven attributes.
The form is filled step by step indicating which step a user is on and explaining what is required as per step.

## Discussion

The earlier discussion was about trying to represent the process of adding an option that gives a similar experience to how
it's done on paper. 
While implementing, I faced an issue with the mat stepper not displaying a form on the different steps, it took some time trying to debug that till I thought of a different way of implementing it naturally where I control the steps through some sort of conditional rendering.

## Preview

https://picsa.app

## Screenshots / Videos
image
image
image
image
image
image
image
image


chrismclarke (changes requested):
Many thanks @khalifan-kfan 

Overall the functionality is definitely heading in the right way, I'm particularly impressed to see you have already started thinking about linking the data with the prompts (great to see the responses stored in the table). 

I'll add a few minor comments inline, but otherwise I think my only change requested will be that you take a look at the linting check to try and make sure it passes. 

**Linting**
Please could you tidy the code to fix the warnings detected in the lint process. Ideally you should see them in your code editor underlined, and with comments if you hover, e.g. 
image

If you do not seem them you may need to install the [eslint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)

Alternatively you can run the lint command for the nx workspace
image

This will show the full output for the lint script (currently 3 errors, 11 warnings identified)
image

apps/picsa-tools/option-tool/src/app/components/editor/editor.component.ts
```diff
@@ -1,8 +1,91 @@
...
+  }
+  handlePracticeEntryNext(){
+    this.stepCounter +=1; 
+  }
+  handlePrevious(){
+    if((this.stepCounter-1) > 0){
+      this.stepCounter -=1;
+    }
+  }
+  handleGender(gender:string){
```
chrismclarke, line 44: + thanks for separating all of these into smaller functions. It will make it much easier if we decide to refactor into separate component files in the future

apps/picsa-tools/option-tool/src/app/material.module.ts
```diff
@@ -1,10 +1,17 @@
...
 import { MatDialogModule } from '@angular/material/dialog';
 import { MatTableModule } from '@angular/material/table';
 
-const COMPONENTS = [MatButtonModule, MatDialogModule, MatTableModule];
+const COMPONENTS = [
+  MatButtonModule,
+  MatDialogModule,
+  MatTableModule,
+  FormsModule,
+  ReactiveFormsModule,
```
chrismclarke, line 13: comment(blocking)
I'm glad to see you figured out how to use angular forms with the use of the FormsModule and ReactiveFormsModule. 
As these imports don't strictly relate to the material components I think we can probably move to the main `app.module` instead, which will still inherit in your template

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
```diff
@@ -25,15 +23,28 @@ const COLUMNS = [
   },
 ];
 export interface IOptionData {
-  name: string;
-  position: number;
-  weight: number;
-  symbol: string;
+  practice: string;
+  gender: string [];
+  benefits: {benefit:string, beneficiary:string[]} [];
```
chrismclarke, line 28: (+)
Appreciate you thinking about these data structures, and that fact that there can be multiple beneficiaries to each benefit

chrismclarke:
I've also added some suggested next development issues for after this is merged in https://github.com/e-picsa/picsa-apps/issues/106

khalifan-kfan:
thanks for the review @chrismclarke, definitely going to work on the feedback

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
khalifan-kfan: thanks

apps/picsa-tools/option-tool/src/app/components/editor/editor.component.ts
khalifan-kfan: would like to see how we approach this

khalifan-kfan:
Hello @chrismclarke fixed the linting, thank you

chrismclarke (approved):
Thanks @khalifan-kfan , I can see you addressed all the feedback comments and now the linting is also passing so I'd say good to go. Feel free to pick up whatever you want to work on next from the list in #106

Issue #106, opened 2023-02-17
[Checklist #3] Options analysis component


chrismclarke:
## Next Priorities
- [x] Ability to export as image to share #238
- [x] Separate options lists for livestock and livelihood options #255 
---

## Feedback TODO
**Updates 2023-09**
- [ ] Fix bug where adding new option does not always appear in expected order #255

**2023-03 UoR feedback**
- [ ] Will need a careful design for direct to farmer App
- [ ] Extension_farmer app add information about pests and diseases and market information as additional columns to matrix

---
## Feedback Completed
- [x] Ability to record multiple sets of options, e.g. initial screen and separate pages for crop, livestock or livelihood option lists
- [x] Ability to go direct to stepper section when clicking on row, e.g. clicking on the _Time_ entry opens stepper on step 7
- [x] Allow users to edit their input by tapping on the same row
- [x] Add support for icons
- [x] Incorporate a mat-stepper to make it easier for users to go back and forth between pages #108 
- [x] Add risks section to stepper (seems to be missing)
- [x] Auto-save on every edit (instead of manual save at end or using save changes button, even if data incomplete)
- [x] Add support for translations (https://docs.picsa.app/translations) #127 
- [x] Save to database #128
- [x] Add page to view and edit all saved lists of options
- [x] Integrate into main app #120

---

**Additional context**
Icons will require import of the `MatIconModule`. which should allow inclusion of any standard material icons. You should be able to test this with an example, e.g. `<mat-icon>home</mat-icon>`

In order to use custom icons you will need to add svg files into the assets folder, and register them in the same way as `apps\picsa-tools\budget-tool\src\app\app.component.ts`. Icons can then be imported with syntax `<mat-icon svgIcon="picsa_icon_name"></mat-icon>
More info in docs https://material.angular.io/components/icon/overview

chrismclarke:
Tracking for specific app features and improvements has been moved to [E-PICSA Issue Triage](https://docs.google.com/spreadsheets/d/17OqUtZZO5aznegT1TSGPtdYOhmtfhBYycgIYx7z0LTs/edit?usp=sharing) Google Sheet instead. Closing here

PR #108, opened 2023-03-08
add mat stepper

khalifan-kfan:
## Description

This Pull request is to modify the option tool by adding a mat-stepper in the case of imputting a new option

## Discussion

 Adding matstepper as requested from issue #106 

## Preview

https://picsa.app

## Screenshots / Videos
image
image
image
image
image
image
image


apps/picsa-tools/option-tool/src/app/components/editor/editor.component.ts
```diff
@@ -32,15 +34,15 @@ export class EditorComponent implements OnInit {
   ]
   this.perfomanceValues = { lowRf:"ok",midRf:"ok",highRf:"ok"}
   this.investmentValues = {money:'high', time:'high'}
+  this.warningText='';
```
chrismclarke, line 37: nit(non-blocking)
As these values are hardcoded you can also just provide the default values when you declare the variable above, i.e. 
```
export class EditorComponent implements OnInit {
  // stepCounter: number;
  warningText = ''
```
The benefit of this is less duplication, and you do not need to explicitly define the type as it will be inferred from the deafult value (will assume string)

chrismclarke (approved):
Thanks @khalifan-kfan, nice job getting the overlay animation working smoothly, the mat-stepper is a nice UI improvement for sure also.

All looks good to me, I've added one minor non-blocking comment but will go ahead and merge here (feel free to address in future PR).

As for next tasks, I think there should still be a couple items on the list (namely making editable and use of icons). Thanks again, good work!

PR #117, opened 2023-03-23
Enable editing of a single option row

khalifan-kfan:
## Description

Changes in this pull request enable a user to edit data in a row and save their changes.

## Discussion

This is to handle an issue that was set out on the option-tool 
Issue:  https://github.com/e-picsa/picsa-apps/pull/103

## Preview

https://picsa.app

## Screenshots / Videos
image


apps/picsa-tools/option-tool/src/app/components/editor/editor.component.ts
```diff
@@ -1,6 +1,15 @@
 import {Component,EventEmitter, OnInit, Output ,ViewChild} from '@angular/core';
 import { MatStepper } from '@angular/material/stepper'
 
+export interface IOptionData {
```
chrismclarke, line 4: Nice idea to store everything in a single data object, I think it works well and makes a lot of sense to me

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
```diff
@@ -29,11 +33,30 @@ export class HomeComponent {
...
     this.matStepperOpen = false
   }
-  onDataTransfer(data: IOptionData) {
+  onDataTransfer(data:{data:IOptionData,index:number}) {
     console.log('Received data from editor:', data);
     // close the child component
-    this.dataSource = [...this.dataSource, data];
+    if(data.index > -1){
+    this.dataSource[data.index]=data.data
+    //invoke change detaction
```
chrismclarke, line 41: nice catch, yes change detection isn't triggered when nested inside an object or array.

One option like you have used is to recreate the entire array, the other would be to manually call angular's change detection method as outlined in https://angular.io/api/core/ChangeDetectorRef

For now I think you are fine with the method you have chosen, so more just sharing in case of interest

chrismclarke (approved):
Works well for me, I've added a couple non-blocking comments in-line but overall I think good to go and a nice addition to have, many thanks and good job.

I've updated the list in #106 to include some next tasks for this feature whenever you have the time

apps/picsa-tools/option-tool/src/app/components/editor/editor.component.ts
khalifan-kfan: Yes sure that would be better

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
khalifan-kfan: Alright thank you

PR #118, opened 2023-03-27
Add some icons to mat stepper

khalifan-kfan:
## Description

Add a few icons to thee map stepper in the option tool

## Discussion

Tracked in issue #103 
## Preview

https://picsa.app

## Screenshots / Videos

image
image
image
image
image
image


chrismclarke (changes requested):
Thanks for adding @khalifan-kfan - I like how you've used a bunch of the existing mat-icons where possible and nice idea to try and add custom img tags to include our own svgs.

As mentioned on slack it would be good to refactor the img tags to be custom mat-icons instead, as that way we can more easily apply consistent styling and alignment. But otherwise I think we have a really nice base to demo and handover to research and design teams for final specifications

khalifan-kfan:
Oh thank you @chrismclarke, I will definitely  modify that. And ping you when I push the changes

apps/picsa-tools/option-tool/src/app/app.component.ts
```diff
@@ -7,4 +10,27 @@ import { Component } from '@angular/core';
...
+  }
+   // register custom icons from the assets/svgs folder for access within the app
+  // icons can be accessed in mat-icon as svgIcon='station_data_${key}'
+  registerIcons() {
+    const icons = {
+      female: 'female',
+      male: 'male'
+    };
+    for (const [key, value] of Object.entries(icons)) {
+      const iconName = `picsa_options_${key}`;
```
chrismclarke, line 28: nice idea to add a custom prefix to the icons, will make them easier to identify in the source code

apps/picsa-tools/option-tool/src/app/components/editor/editor.component.html
```diff
@@ -1,198 +1,208 @@
...
+
+    <mat-step matStepContent label="Gender ">
+      <div class="StepContainer">
+        <div class="StepTitle">Who does the practice?</div>
+        <div class="StepContent">
+          <div class="genderDiv">
+            <div class="genderType" style="color: #800080;" (click)="handleGender('female')"
+              [ngStyle]="{'background-color': gender.includes('female') ? 'pink' : 'lightgray'}">Female 
+              <!-- <img
+                *ngIf="femaleIcon" [src]="femaleIcon" /> -->
```
chrismclarke, line 31: nit(non-blocking)
If the code is no longer needed probably cleaner to just remove, git keeps code history if you ever need to recover. There's a few comments in this file that could probably be deleted

chrismclarke (approved):
Looks good to me, I've included a couple minor non-blocking comments so these can just be resolved during future work. Nice job!

PR #121, opened 2023-04-13
Ft Crop-information-tool

FaithDaka:
## Description
- Includes initial crop probability table design

## Discussion
The table filter is currently commented due to display bugs. When the filter is active, the probability table doesn't show and the CSS is distorted.

## Screenshots / Videos

image


apps/picsa-tools/crop-probability-tool/src/app/components/crop-probability-table-header/crop-probability-table-header.component.html
```diff
@@ -0,0 +1,24 @@
+<table ngClass="header-table sticky">
+  <thead>
+    <tr ngClass="header-title fixed-header">
+        <th colspan="4" rowspan="4">Crop Information Sheet – CHIPATA MET STATION </th>
+    </tr>
+    <tr ngClass="right-cells fixed-width">
+        <th>1-Nov</th>
```
chrismclarke, line 7: nit(non-blocking)
Likely at some point these headings will also come from the data definition, but fine to hardcode for now

apps/picsa-tools/crop-probability-tool/src/app/components/crop-probability-table-header/crop-probability-table-header.component.ts
```diff
@@ -0,0 +1,8 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'picsa-crop-probability-table-header',
+  templateUrl: './crop-probability-table-header.component.html',
+  styleUrls: ['./crop-probability-table-header.component.scss'],
+})
+export class CropProbabilityTableHeaderComponent {}
```
chrismclarke, line 8: nice idea to split this into it's own component, definitely helps keep the rest of the code a lot tidier

apps/picsa-tools/crop-probability-tool/src/app/components/crop-probability-table/crop-probability-table.component.html
```diff
@@ -1,43 +1,85 @@
...
-  <mat-icon>menu</mat-icon>
-</button>
-<table mat-table [dataSource]="dataSource" class="mat-elevation-z8">
-  <!--- Note that these columns can be defined in any order.
-        The actual rendered columns are set as a property on the row definition" -->
+<!-- <div ngClass="table-filter">
+  <mat-form-field>
+    <mat-label>Filter</mat-label>
+    <input
+      matInput
```
chrismclarke, line 5: I can see this is the implementation you had issues with. Taking a quick look your code is good, I think there's just a minor issue because the `matInput` you've referenced hasn't been imported into the material module 

(super annoying I know, so many separate imports and the error message it not even remotely useful - I only realised by chance)

I'll add a commit to include

apps/picsa-tools/crop-probability-tool/src/app/data/mock.ts
```diff
@@ -1,23 +1,169 @@
...
-    variety: 'SC 637',
-    days: '130-136',
-    water: '512-535mm',
-    probabilities: ['10/10', '10/10', '8-9/10', '3-5/10', '0-1/10'],
+    data: [
+      {
+        variety: 'SC 637',
+        days: '130-136',
+        water: ['512-535mm'],
+        probabilities: ['10/10', '10/10', '8-9/10', '3-5/10', '0-1/10'],
```
chrismclarke, line 9: Really nice having such a comprehensive example, thanks for adding all of this

chrismclarke (approved):
Really nice work! Very impressed to see all the the different cell merging working, I imagine it must have been quite tricky to work out

I was also able to get your filter working, which is really nice to have also, so I think to start with this is all good to go. I'll update the checklist for potential next tasks, but for now thanks for the addition!

PR #122, opened 2023-04-14
Feat/option tool autosave

khalifan-kfan:
## Description

This allows the user to save midway through the process of filling the stepper. This comes in handy if the user doesn't want to finish the process but saves what they have so far filled in the stepper.
Also, this PR refactors the logic of the buttons just slightly.
Not sure if "autosave" is the right word to describe this PR though.  :-|

## Discussion

This feature follows the discussion from the option tool tasks checklist
[#106]
Next will be the integration of translation

## Preview

https://picsa.app

## Screenshots / Videos
image
If a user tries to save without giving  at least the name
image


chrismclarke (approved):
Thanks @khalifan-kfan 
I tested the functionality and working well, nice work.

I've made a few edits to add functionality to delete a saved entry from the practice page, and made a few of the buttons more consistent with the budget tool. It seems my code editor made a lot of additional edits, likely due to the prettier plugin I have installed (e.g. swapping double quotes for single), so the changes look much bigger than they are.

I'll try organise a follow-up PR to enforce consistent code styling automatically so there aren't such big diffs when I try to make edits, but for now I'd say all this is looking good and moving onto translations sounds like a solid idea

PR #127, opened 2023-04-26
Feat/option tool transilation

khalifan-kfan:
## Description

Add  translation feature to option tool

## Discussion

This follows the tasks related to the option tool analysis

## Preview

https://picsa.app

## Screenshots / Videos
For some pre-existing translations
image


chrismclarke (approved):
Thanks @khalifan-kfan , this looks perfect to me. I know it's hard to test, as we will only see the populated translations after we have sent our updated list of translations to translators and included in the main list, but looking at the generated `libs/i18n/_template.csv` file I an see most of the text I would expect which shows you have configured correctly.

I've added a couple minor PRs on top to include automatically spellchecking generated translations and fixed 1 minor typo detected. I've also added some common translations so that when we have repeats for works like "Close" and "Back" the jsons will not update (still fine as all apps have access to all translations, just makes it easier for tracking diffs on github)

But otherwise looks really good to me, many thanks for the contribution

PR #130, opened 2023-05-04
Feat(crop-probability): Filter Icons

FaithDaka:
## Description

- Filter crop probability table by crop icons

## Discussion
Understanding the routing in the app was complicated for me esp for instances like the _Resources_ app.

The dead-ends I tried in order to route to crop-probability were:
  - Creating a resource type for components that required routing, this was added to the resources _model.ts_ file. Then I created a component resource in the _crop_ data file to be rendered as a child of the _Crop, Pest and Disease_ resource type. But I think those are auto generated using a certain command since each had a _key and _timefields for creation and modification. This would have made it easy to route to the component on click.
  - Also tried "hard-routing" through the resources page but there were complexities due to how the resource items are represented.
 
For the crop icons:
- When rendering the custom icons, the assets URL is not appended correctly. Is that attributed to the routing implemented?  
image

## Screenshots 
image


FaithDaka:
A few discussion points @chrismclarke

chrismclarke (approved):
Thanks @FaithDaka 
Following the call I took a quick look at how image assets are handled and refactored the budget card images to make them easier to share with the crop-probability tool. I also updated the imports to include the shared assets and removed the ones you had added, all in commit 0cd1ae4

So I think things should now display as expected

localhost_4200_crop-probability(PICSA)

I've also added documentation to the site to explain how angular shares assets across different tools/apps, because it's pretty confusing at the best of time (hope the docs help it make sense a bite more)
https://docs.picsa.app/advanced/assets#sharing-across-tools

As for next tasks, I think the easiest way to address routing for now would be within the tool itself (e.g. create a dropdown for site name/id, and use that to change the data loaded). 

I'll be happy to help extract that functionality to link from the resources page in the future, although I first want to check with the university team if they would prefer to link from resources or if we are going to be implementing new home screen designs that might be a better place to link from.

The other task required would be to add support for translations. There should be some documentation at https://docs.picsa.app/translations, but feel free to ask either me or Khalifan if you have any questions about it (as he recently also implemented for the option tool)

Otherwise nice to see things moving forward, thanks again

apps/picsa-tools/crop-probability-tool/src/app/components/crop-probability-table/crop-probability-table.component.html
```diff
@@ -1,8 +1,17 @@
-<div ngClass="table-filter">
+<!-- <div ngClass="table-filter">
   <mat-form-field>
     <mat-label>Filter</mat-label>
     <input matInput (keyup)="applyFilter($event)" placeholder="Search Crops" #input />
   </mat-form-field>
+</div> -->
+<div ngClass="filter-icons">
+  <div (click)="filterbyImage('maize')"><mat-icon svgIcon="picsa_crop_maize"></mat-icon><div>Maize</div></div>
```
chrismclarke, line 8: nit(non-blocking) - it might be better to refactor this to an `ngFor` loop, with a list of crops in the component .ts file. If stored as a json object you could also track the label and svgIcon name

FaithDaka:
Alright, thanks for the detail @chrismclarke 
Will be helpful moving forward

PR #131, opened 2023-05-15
Feat/options rxdb connection

khalifan-kfan:
## Description

This PR integrates the option tool to an in browser indexed database. It allows a user to update, delete and edit data

## Discussion

   #106

## Preview

https://picsa.app/option

## Screenshots / Videos

image


chrismclarke (approved):
Thanks @khalifan-kfan , lots of really useful code here and super nice to see rxdb integrated and working smoothly. In particular I really like how you managed to find a way to initialise the async db creation, and provide strong type defintions to use with the db queries

I've added 2 follow-up commits to help address the challenge of turning this into a feature that can be used across other tools also, specifically

5a50f57
Creates a new shared module that can be used to import the database service into any module. It uses a special `forRoot` pattern which avoids duplicate initialisation across imports (there will always be just a single global provider, whether using the option-tool standalone, or alongside multiple other tools integrated into the extension app)

The angular docs covers this concept in more detail at: 
https://angular.io/guide/singleton-services

196c506
Creates a new option-tool service to refactor some of the code used in the home component,, remove some duplicate type definitions, and change way data is queried.
I've opted to use rxdb's live-query feature, that allows a component to query the database and receive realtime updates when data changes (instead of manually calling a refresh function).

This pattern is pretty common in angular too, although has one minor caveat that any subscriptions need to be unsubscribed when the component is destroyed as can otherwise lead to memory leaks. The article below covers a few ways to do this (you'll see I opted for the rxjs takeUntil utility). 

Let me know if you have any questions about the changes, otherwise I think looks good to go!

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
```diff
@@ -1,60 +1,84 @@
...
-export interface IOptionData {
-  practice: string;
-  gender: string[];
-  benefits: { benefit: string; beneficiary: string[] }[];
-  performance: { lowRf: string; midRf: string; highRf: string };
-  investment: { money: string; time: string };
-  time: string;
-  risk: string;
-}
+import { IOptionsToolEntry } from '../../schemas/schema_v0';
```
khalifan-kfan, line 7: Nice one

Issue #138, opened 2023-06-01
Feat(manual) - Add Interactive manual for direct-to-farmer version


chrismclarke:
**Dev Priorities**
- [x] Add new tool to store interactive manual (future direct-to-farmer home page) #139
- [x] Initial designs 
- [x] Placeholder sections for all PICSA steps
- [x] Linked pages for step content and activities #142
- [x] Integrate video player (e.g. https://www.npmjs.com/package/capacitor-video-player)
- [ ] Example manual content rendered from markdown
- [ ] Create standalone repo to store full manual in markdown format with translations

**Additional Links**
[Initial Designs](https://www.figma.com/file/SSD66rh7cUx2deNXZDl6Gl/E-PICSA-app-LAYOUT?type=design&node-id=361-219&t=j3oF03YlBt5Wv4JG-0)
[Dev Meeting Notes](https://docs.google.com/document/d/1IwnmdQDsnMfiwC7ln8MxU3AnsUJZODTMZhczTRTM964/edit#heading=h.ei82jf5l8a7q)

PR #142, opened 2023-06-25
Add routing for manual activity

khalifan-kfan:
## Description

In grid view of manual items when one clicks an activity, they should be routed to the activity details page

## Discussion

Following the discussion from #138

## Preview

https://picsa.app

## Screenshots / Videos

Screenshot 2023-06-25 at 16 48 14


apps/picsa-tools/manual-tool/src/app/pages/activity-details/activity-details.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  templateUrl: './activity-details.component.html',
+  styleUrls: ['./activity-details.component.scss'],
+})
+export class ActivityDetailsComponent implements OnInit {
+  activity: IManualActivity;
+
+  constructor(private route: ActivatedRoute) {}
+
+  ngOnInit() {
+    this.activity = history.state.activity;
```
chrismclarke, line 17: nit(non-blocking)
Passing the activity using history state will mean the page will only work when navigated to via link, but it might also be nice to work directly (e.g. someone shares a url directly to the activity page)
A better way to implement might be to use the route `paramMap` to identify the activity id, and retrieve from the list of all activities

chrismclarke (approved):
All looks good, I've added a couple small tidying commits

d53c6bc - add stronger type definitions for activities
eeda5c7 - add activity id router lookup

apps/picsa-tools/manual-tool/src/app/pages/activity-details/activity-details.component.ts
khalifan-kfan: Yes I get your point I hadn't thought of that use case

khalifan-kfan:
hey @chrismclarke, so I was intending to open another PR more focused on the details per activity because for starters I wanted to first route to that Page (screen short) then after determine what exactly to display on each ativity when the buttons are clicked. Also I was trying to follow the activity check boxes

khalifan-kfan:
Sorry @chrismclarke I forgot to push the last commit

PR #143, opened 2023-06-27
Add activity details page

khalifan-kfan:
## Description

In the grid view of manual items when one clicks an activity, they should be routed to the activity details page
@chrismclarke This is to cover for  the forgotten commit, so sorry about that, It is also merged with your code tiding

## Discussion

Following the discussion from #138

## Preview

https://picsa.app

## Screenshots / Videos

Screenshot 2023-06-25 at 16 48 14


chrismclarke (approved):
Thanks @khalifan-kfan , all looks good to me.
I appreciate the placeholder video too, definitely makes me thing we will need to be careful about any videos we include so the app doesn't get too big (even the 10s countdown video adds 1mb to the main build, bumping from around 15MB -> 16MB)

I've compressed the demo video using h.265 video encoding which produces slightly smaller file sizes, and reduced the frame size to 720p which reduces the overall size about 50% - I'll be interested to test if this still looks good on mobile and if yes will add some documentation to explain how best to optimise videos. 

Otherwise I think the next task metioned in #138 will be to test integration of the capacitor video player to see if that also provides a better video viewing experience on native

khalifan-kfan:
Sure, I'll try out the library on the next task and give you feedback as well, thanks for compressing the video

khalifan-kfan:
If the video sizes later become very huge, we could parse them through links and host them online

PR #149, opened 2023-07-03
Feat/New stations

FaithDaka:
## Description

- New MET stations added for the crop probability tool i.e petauke and kasungu

## Screenshots

image


FaithDaka:
Ready for review @chrismclarke

chrismclarke (approved):
Thanks @FaithDaka - really appreciate some of the new stations had a lot of data to capture so thanks for being really diligent. I've done my best to cross-check all the sites, stations, and a selection of crops and all looking good to me so I'd say good to go.

I think we can also go ahead and remove the hardcoded PDF resources which frees up almost 1MB from the core build which is very nice to have (will add a commit to this PR)

PR #203, opened 2023-12-08
feat(seasonal-calendar): add translations

khalifan-kfan:
## Description

This PR is to enable translation on the seasonal calendar tool

## Discussion

Building the seasonal calender


chrismclarke (changes requested):
Thanks @khalifan-kfan 

I noticed a typo in the scripts that extracts translations which explains why they weren't generating as expected, so I've fixed and rerun the extraction process `yarn:i18n` and now can see correctly the additional strings marked for translation

image

With the additional translations extracted, it is now possible to verify what strings have been marked for translation from the frontend by using the `Translators` preview language select option

image

When looking at the seasonal-calendar home page I can see all strings correctly have the markers indicating that they are linked up to translations

image

However when I try to create a new seasonal calendar I hit an error and blank page (likely an issue in the html code, somewhere you are using text `85}}`)

image

So when you get a chance can you fix the error, and continue in the same way to check whether all translations are marked as expected, and if not try to get them included.

Any new translations marked will appear in the list after you rerun the `yarn i18n` script

khalifan-kfan:
Sure thanks @chrismclarke , I am going to work on this feedback

khalifan-kfan:
Hello @chrismclarke, thanks for the feedback and help. 
All changes have been worked on.

chrismclarke:
Thanks @khalifan-kfan !

I found a few minor strings without translation (e.g. some of the error messages, the months dropdown), so have updated them in 
efd8ad7 and ae33619

image

I also noticed it wasn't possible to translate the crop names on the table as the table essentially populates a crop id and not the label. I think probably easiest if we merge this PR here and I'll try to add a follow-up with any required data schema and binding changes. Thanks again for making the changes

chrismclarke (approved):

Issue #208, opened 2023-12-14
feat(dashboard): translation management


chrismclarke:
**Is your feature request related to a problem? Please describe.**
Currently all translations are managed using google sheets, which requires multiple manual steps to integrate into the app. Ideally a minimal translations management system should be included in the admin dashboard

**Describe the solution you'd like**
- Create translations page in dashboard
- Create db table with columns for id and translations by country/language codes (e.g. `en`, `mw_ny`, `zm_ny`)
- Create user interface to create and edit translations

**Additional context**
Add any other context or screenshots about the request, or relevant parts of the codebase if known

PR #211, opened 2024-01-03
Ft server translations

khalifan-kfan:
## Description

Enable display and editing of translations on the dashboard

## Discussion

issue [#208](https://github.com/e-picsa/picsa-apps/issues/208)

## Preview

## Screenshots / Videos

Screenshot 2024-01-03 at 09 24 01


apps/picsa-apps/dashboard/src/app/modules/translations/pages/edit/translations-edit.component.html
```diff
@@ -0,0 +1,47 @@
+<div class="page-content">
+  <h2>Edit Translations</h2>
+  @if(translationRow){
+  <form class="form-content">
+    <div class="form-data">
+      <label>Date of Creation</label>
+      <div>{{ translationRow.created_at | date: 'mediumDate' }}</div>
```
chrismclarke, line 7: (+) - Nice use of the date pipe to display a friendlier format, I had forgotten all about that one!

apps/picsa-apps/dashboard/src/app/modules/translations/pages/edit/translations-edit.component.html
```diff
@@ -0,0 +1,47 @@
...
+  <div>{{dataLoadError}}</div>
+  } @else {
+  <div>Loading...</div>
+  }
+  @if(editActionFeedbackMessage){
+  <div>{{editActionFeedbackMessage}}</div>
+  }
+  @if(translationRow){
+  <h2>Delete this Translation</h2>
+  <button mat-raised-button color="primary" class="deleteButton" (click)="openTranslationDeleteDialog()">Delete this
```
chrismclarke, line 44: nit(non-blocking) - Really nice to experiment with all the possible functionality to add/remove translations. Once deployed to production I think probably we will remove the option for deleting translations (as the list will be automatically generated from the app and not user-controlled), but will keep for now to make things easier to test.

apps/picsa-apps/dashboard/src/app/modules/translations/translations.service.ts
```diff
@@ -0,0 +1,86 @@
...
+    const { data, error } = await this.supabaseService.db.table('translations').select<'*', ITranslationRow>('*');
+    if (error) {
+      throw error;
+    }
+    this.translations = data || [];
+  }
+   
+  // update a translation record by ID
+  public async updateTranslationById(id: number, updatedData: Partial<ITranslationRow>): Promise<string> {
+    const { data, error } = await this.supabaseService.db.table('translations').update(updatedData).eq('id', id);
```
chrismclarke, line 44: nit(non-blocking) - up till now the way ids have been generated for app translations is a combination of the text in english along with 2 metadata columnns titled `tool` and `context`. This means that we avoid duplicate translations if something like the word "close" appears multiple times (e.g. lots of tools have dialogs with a close button), but allows for multiple translations where required (e.g. "close" dialog vs the distance is not far but is "close").

So what this means in practical terms is instead of searching for an id we will need to search for a composite match of en text with context and tool data fields. 

Fortunately supabase supports this via chaining query syntax, e.g. `.eq('en', en).eq('tool',tool).eq('context',context)`

However I don't think this is need right now, as I think possible we will want to refactor to remove the edit page and allow users to directly edit on the main page (possibly from the table itself) - so suggest no further action required right now but just wanted to share the information. Hope that makes sense

apps/picsa-apps/dashboard/src/app/modules/translations/pages/edit/translations-edit.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  templateUrl: './translations-edit.component.html',
+  styleUrls: ['./translations-edit.component.scss'],
+})
+export class TranslationsEditComponent {
+  translationRow: ITranslationEntry;
+  dataLoadError: string;
+  editActionFeedbackMessage: string;
+  constructor(private service: TranslationDashboardService, private route: ActivatedRoute, private router: Router) {
+    this.service.ready();
+    this.route.params.subscribe((params) => {
```
chrismclarke, line 28: nit(non-blocking) - nice implementation to track route params. Just 2 minor issues:

1. Any time you create a subscription you are using browser memory. So when the subscription is no longer required you should remove it. This is commonly done by linking the subscription to ngOnInit and ngOnDestroy methods, e.g. 

2. I've actually broken this implementation/page due to changes to how the `id` field is calculated. I'd say no point fixing right now as we will want to first decide what will be the best user flow for updating translations

apps/picsa-apps/dashboard/src/app/modules/translations/pages/home/translations.page.html
```diff
@@ -0,0 +1,61 @@
+<div class="page-content">
+  <div style="display: flex; align-items: center">
+    <h2 style="flex: 1">Translations</h2>
+    <button mat-raised-button color="primary" routerLink="/translations/new">Add New Translation</button>
+  </div>
+  @if(service.translations){
+  <table mat-table [dataSource]="service.translations">
```
chrismclarke, line 7: (+1) Nice to see the mat-table implementation. I'm currently creating a shared component to make it easier to use mat tables as always feels like a lot of boilerplate code to have to write. But think this works well for now so I'll just let you know when the shared component is ready.

apps/picsa-apps/dashboard/src/app/modules/translations/pages/new/new-translations.component.html
```diff
@@ -0,0 +1,34 @@
...
+      <input matInput [(ngModel)]="tg" name="tg" />
+    </div>
+    <div class="form-data">
+      <label>zm_ny</label>
+      <input matInput [(ngModel)]="zm_ny" name="zm_ny" />
+    </div>
+    <button mat-raised-button color="primary" class="submitButton" (click)="submitForm()">Submit</button>
+  </form>
+
+  @if(editActionFeedbackMessage){
```
chrismclarke, line 31: (+1) - nice to see you are using the new angular control flow syntax. Personally I find the html markup is a lot cleaner with it, so will probably suggest we keep preferring to use it over the older ngIf/ngFor version

apps/picsa-server/supabase/migrations/20240102153452_translations_table_create.sql
```diff
@@ -0,0 +1,17 @@
...
+    -- Uses ISO 3166 alpha-2 country code and  ISO 639 Set 1 language code
+    -- https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
+    -- https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes
+    mw_ny text null,
+    ke_sw text null,
+    tj_tg text null,
+    zm_ny text null,
+    -- Primary key generated as combination of tool, context and en text
+    constraint translations_pkey primary key (tool, context, en)
+  ) tablespace pg_default;
```
chrismclarke, line 17: (+) - Thanks for exporting the defintion, can confirm very smooth for me to recreate the table locally using this.

I've made a few small tweaks, to include a couple extra metadata columns and to change the way the primary key is generated. See update in commit 813935b

apps/picsa-server/supabase/types/index.ts
```diff
@@ -264,6 +264,36 @@ export interface Database {
...
+      translations: {
+        Row: {
+          created_at: string
+          en: string | null
+          id: number
+          mw_ny: string | null
+          sw: string | null
+          tg: string | null
+          zm_ny: string | null
+        }
```
chrismclarke, line 276: (+) - Thanks for remembering to export the definitions, makes it really nice to be able to write type-safe code in the services and display component

libs/shared/src/services/core/supabase/components/storage-file-picker/storage-file-picker.component.ts
```diff
@@ -66,7 +66,7 @@ export class SupabaseStorageFilePickerComponent {
     const entries = await this.storage.list(this.storageBucketName, this.storageFolderPath);
     this.fileEntries = entries
       // filter out metadata files which have filename starting with `.`
-      .filter(({ name }) => !name?.split('/').pop()?.startsWith('.'))
+      .filter(({ name }) => name && !name.split('/').pop()?.startsWith('.'))
```
chrismclarke, line 69: (+) - thanks for tidying this

chrismclarke (approved):

chrismclarke (approved):
Really nice implementation overall, great to see you were able to implement pretty much all aspects of backend server and dashboard (creating tables, generating types, setting up all necessary db bindings).

You'll see I've added one commit to update the db columns - as I've made the changes directly (instead of writing a migration) you'll need to reset your local db to update the changes `yarn nx run picsa-server:supabase db reset`. You'll also see I've included some initial data in the `seed` folder, which can be imported locally using the supabase dashboard (there's a button to import csv when you load an empty table)

As part of next tasks I think we will probably want to decide what will be the best way to allow administrators to update translations (and how to update the main list from the app), but for now I think looks really good and should provide a solid base to build from so nice work!

chrismclarke:
@khalifan-kfan - Minor update to follow previous comments.

I tested implementing my proposed changes to how ids are calculated (generating at the db level by combining tool, context and en columns), however this makes all other navigation tricky because there's no longer an `id` query parameter to search for.

So instead I've re-implemented to keep an `id` field in the database which is still generated as a string combination of the fields, specifically by converting the en text to lower case without spaces and special characters, and combining with the `tool` and `context` columns
```ts
 const textId = row.en.toLowerCase().replace(/[^a-z0-9]/gi, '');
    return `${row.tool}-${row.context || ''}-${textId}`;
```    

I've updated the db definition and seed data accordingly, so hopefully should be clearer once imported.

apps/picsa-apps/dashboard/src/app/modules/translations/translations.service.ts
khalifan-kfan: Thanks for this.

apps/picsa-apps/dashboard/src/app/modules/translations/pages/edit/translations-edit.component.ts
khalifan-kfan: Thanks for this, will surely tidy up next time

apps/picsa-apps/dashboard/src/app/modules/translations/pages/new/new-translations.component.html
khalifan-kfan: I enjoyed using it too

khalifan-kfan:
Thanks for the feedback and updates on this PR.

Issue #218, opened 2024-01-18
feat(dashboard) - Monitoring summary


chrismclarke:
**Is your feature request related to a problem? Please describe.**
We want to make it easier for administrators to manage monitoring forms. In the short term this will just be read-only access to lists of forms and submissions (will later look at ways to add new forms)

**Describe the solution you'd like**

_Form summary_
- [x] Create dashboard pages for `/monitoring` 
- [x] Create db endpoint for `monitoring_forms` with schema matching hardcoded data found in `apps\picsa-tools\monitoring-tool\data\forms`
- [x] Manually populate seed data for the 3 hardcoded forms. Once populated export as csv and place in the `apps\picsa-server\supabase\data` folder to make it easier for others to import
- [x] Create db bindings and UI to list forms and display in a table

_Form submissions_
- [x] Create page for `/monitoring/{formId}` (using route param for form id)
- [x] There is already a db schema for `monitoring_tool_submissions` so I think fine to leave as-is for now
- [x] Create db bindings to get all submissions by a form id and display in a table

**Additional context**
1) A new `picsa-data-table` component has been added to make it easier to generate simple data tables (uses mat-table but less boilerplate code required). This can be imported into any page, and should be used to create the UI tables instead of manually creating new mat-tables.

2) The easiest way to generate seed data for `monitoring_tool_submissions` is to run the app locally, fill in a form and submit. It should automatically sync to the local supabase db.

![image](https://github.com/e-picsa/picsa-apps/assets/10515065/3887b8d6-7755-40a6-81b9-e3d48871ab9f)

![image](https://github.com/e-picsa/picsa-apps/assets/10515065/b8340cb4-31c2-44cb-a046-5bdf9859e507)

PR #231, opened 2024-02-13
feat(dashboard): add monitoring pages and forms db

khalifan-kfan:
## Screenshots (updated)
localhost_4200_monitoring(720p)
localhost_4200_monitoring(720p) (1)
localhost_4200_monitoring(720p) (2)

---
## Description

Add monitoring forms display pages and database schemas

## Discussion

#218

## Extra information and queries

1. @chrismclarke could you help confirm whether am on the right track and if the schema so far is fine.
   To be more specific,
   1.1 Should I link the cover item to storage by foreign key?
   1.2 For data such as appCountries and appSummary, should I make them relations instead of embedded data in the data 
   base?
   1.3 Have I represented the ```enketoDefinition``` field as required?

 2.I have attempted to use the ```picsa-data-table``` component but as you can see in the screenshot, some fields are presented in a specific type of way, any pointers on what I should put into consideration to effect it?
 

## Screenshots / Videos

Screenshot 2024-02-13 at 14 36 10


chrismclarke:
Thanks @khalifan-kfan , looks like a really good start, let me do my best to address the questions you raised:

> Extra information and queries
> @chrismclarke could you help confirm whether am on the right track and if the schema so far is fine.
> To be more specific,
> 1.1 Should I link the cover item to storage by foreign key?
> 1.2 For data such as appCountries and appSummary, should I make them relations instead of embedded data in the data
> base?
> 1.3 Have I represented the enketoDefinition field as required?

Overall schema looks good, although one minor consideration is to change column names to use `snake_case` instead of `camelCase` (e.g. `enketoDefinition` -> `enketo_definition`). This is purely a matter of convention and isn't super significant, just more common for databases to use snake_case and javascript camelCase. It does then mean some incompatibility if writing the same definitions in both languages, but usually not a big issue as we have automated type generation from the database to use in the code instead of manually typing.

1.1 - I think given that it's an svg actually might be easier to just inline as text and avoid having to integrate storage buckets at this point

1.2 - I think I need to have a more considered think generally about how we're going to manage what data appears in what version of the app. I think for now what you have done is fine and makes sense to just keep as hardcoded per form

1.3 - Yes you have, the field is JSON so using JSONB data type is a good choice. Whilst it would be possible to store as a string, that would add an extra step to parse/stringify when interacting with, and by using JSONB it also makes the data inside the JSON searchable via SQL if we ever wanted to do that


> 2.I have attempted to use the picsa-data-table component but as you can see in the screenshot, some fields are presented in a specific type of way, any pointers on what I should put into consideration to effect it?

Good point, as mentioned on the call I reached the same limitation myself so I've updated the `picsa-data-table` component to provide functionality to override display for a specific cell. I've provided an example by duplicating the table and pushed as commit ee24b72. Let me know if the changes make sense (really I should add some documentation about the component more generally... will put on a todo list). A good challenge would be to see if you can extend the example to format the `created_at` date column nicely, e.g. using the angular [date pipe](https://angular.io/api/common/DatePipe), i.e. `{{value | date}}`

chrismclarke:
I'll revert the PR back to draft for now, but let me know if you have any other questions or clarifications, and when you think next ready for a code review

khalifan-kfan:
Thanks for the response @chrismclarke, should I enable all crud operations to the monitoring forms  table in the UI or I should leave it as read only

chrismclarke:
> Thanks for the response @chrismclarke, should I enable all crud operations to the monitoring forms table in the UI or I should leave it as read only

I think for now probably just leaving readonly - I expect admin would probably like a way to make changes to form submissions in the future, although this will need to be considered a bit more carefully to ensure that any changes admin make don't just get replaced if the user updates the submission from their side (incoming data will always take priority) - so we need a way to ensure any admin changes also get synced back to the user device, and will need to establish a conflict resolution strategy if both an admin and a user try to push differences to the same form submission

khalifan-kfan:
Alright, thanks,
Now for the last part of it 
```Create db bindings to get all submissions by a form id and display in a table```
should this feature only be available on the dashboard.

chrismclarke:
I've removed myself as a reviewer just so I don't get notified as you push commits (annoying github feature), but feel free to tag me when ready for review or if you want me to look at anything else

chrismclarke (approved):
Thanks @khalifan-kfan , I think things are looking really good overall.

I think the setup of form list, form details and form submissions works well, I've made a few minor changes to the url names (shortenting monitoring-tool -> monitoring and replacing numerical `id` with the text `form_device_id` which removes the need for an extra url parameter.

I've made a few minor db schema edits, and renamed the migration so that it is linear with the rest of migrations already merged (always a bit of a pain when there's multiple people working on db migrations, as want to make sure the order they appear are the same as the order they're merged). So it will mean that your local db will be out-of-sync and will require a reset when next working on backend.

You can see the changes I applied in b5c44f1, they look a bit busier than they actually are due to some code auto-formatting (although that shouldn't happen as much in the future as I think I've fixed the tool that handles auto-format for all incoming prs so less developer-dependent).

I'll outline the next steps for the monitoring server in a separate issue so that you can feel free to either continue with it or take on something else depending on interest, but for now thanks for all the work on this, be nice to get merged in

khalifan-kfan:
Thanks for the review @chrismclarke,  let me look at another issue to take on

PR #232, opened 2024-02-13
[Dashboard] Crop Probability Module

FaithDaka:
## Description

- Includes the design of the crop probability schema and table
- Database migration configuration
- Create a crop probability entry

## Discussion

- Pending issues are: returning the nested tables, viewing an individual record

## Screenshots / Videos

image

image


chrismclarke (commented):
Thanks @FaithDaka - looking through the data and changes I can definitely appreciate this is a much more complex set of tasks than originally communicated.

I really like that you have considered both a base set of crop information, and then station-specific changes/overrides. 

I've made some small changes to the `crop_data` table. I learnt it's possible to use multiple columns as the unique id - the benefit in doing this is it restricts the data that can be entered, now so that there is only a single crop-variety entry possible (e.g. `maize-SC304`), along with some minor column edits.
image

I've also reversed the relationship, so now the `crop_data` table has no reference to climate stations, but the `climate_station_crop_data` table can be used to provide a station-specific overrides to default crop information. I think this should make it easier to allow work on crop data independent of station calculations, and it also removes the need for a 3rd `station_data` table linking the two (so now just have two tables). 

image

NOTE - I've applied all schema changes directly on top of existing migrations files, so you will need to reset your local db to re-apply the migration with updated definitions

With the schema updates pushed I'll move onto reviewing frontend code next

apps/picsa-apps/dashboard/src/app/modules/crop-information/crop-information.component.html
```diff
@@ -0,0 +1,36 @@
+<div class="page-content">
+  <div style="display: flex; align-items: center">
+    <h2 style="flex: 1">Crop Probabilities</h2>
+    <button mat-raised-button color="primary" (click)="navigateToEntryPage()">Add New Entry</button>
+  </div>
+  @if(service.cropProbabilities){
+  <table mat-table [dataSource]="service.cropProbabilities">
```
chrismclarke, line 7: nit(non-blocking)
I've added a shared `picsa-data-table` component to reduce the amount of boilerplate needed every time we create a new mat-table. Can you try import from `libs\shared\src\features\data-table\data-table.component.ts` and see if it works as a drop-in replacement for the table code?

apps/picsa-apps/dashboard/src/app/modules/crop-information/crop-information.page.ts
```diff
@@ -0,0 +1,44 @@
...
+    'station_name',
+    'water_lower',
+    'water_upper',
+    'length_lower',
+    'length_upper',
+    'station_notes',
+  ];
+
+  ngOnInit(): void {
+    this.service.ready();
```
chrismclarke, line 31: nit(non-blocking)
The service init code is async so it would be better to `await` the call. Handily enough angular allows converting the `ngOnInit` code to also be async, just simply state as `async ngOnInit(){....}`

apps/picsa-apps/dashboard/src/app/modules/crop-information/crop-information.page.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.refreshCropInformation();
+  }
+
+  refreshCropInformation() {
+    this.service.listCropProbabilities().catch((error) => {
+      console.error('Error fetching crop probabilities:', error);
```
chrismclarke, line 37: nit(non-blocking)
I've added a shared `picsaNotificationService` to make it easier to display error notifications. Could you try to include here, and call with the message passed from the error, e.g. `notificationService.showUserNotification({icon:'error',message:'Error fetching crop probabilities:'+error.message})`

apps/picsa-apps/dashboard/src/app/modules/crop-information/crop-information.page.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+  }
+
+  navigateToEntryPage() {
+    this.router.navigate(['/crop/entry']);
+  }
```
chrismclarke, line 43: nit(non-blocking)
Remembering the previous route prefix when navigating to nested routes can sometimes become a little prone to error (e.g. previous route changes path name). A slightly more resilient solution is to import the activated route in the constructor, and use it to perform relative navigation. I.e.
```ts
constructor(private route: ActivatedRoute)
// use in method
this.router.navigate(['entry'], { relativeTo: this.route });
```

apps/picsa-apps/dashboard/src/app/modules/crop-information/pages/new_entry/new_entry.component.html
```diff
@@ -0,0 +1,35 @@
+<div class="page-content">
+  <button mat-stroked-button color="primary" (click)="routeBack()"><mat-icon> Back </mat-icon></button>
```
chrismclarke, line 2: nit(non-blocking) - appreciate the consideration for managing back navigation, although I think it would probably be better if I implement some sort of breadcrumb/common navigation to use across all pages instead of relying on each page to implement - so can probably just remove here for now and I'll put on my todo list

apps/picsa-apps/dashboard/src/app/modules/crop-information/pages/new_entry/new_entry.page.ts
```diff
@@ -0,0 +1,55 @@
...
+  templateUrl: './new_entry.component.html',
+  styleUrls: ['./new_entry.component.scss'],
+})
+export class NewEntryPageComponent implements OnInit {
+  station_name: string;
+  water_lower: number;
+  water_upper: number;
+  length_lower: number;
+  length_upper: number;
+  station_notes: string[];
```
chrismclarke, line 23: nit(blocking)
It would be really nice if we could manage the data bindings as a single `form` object that can also be used to handle things like validation. 

Could you try implement as a reactive form instead, there should be some good guidance to create the form using the angular formbuilder at: https://angular.dev/guide/forms/reactive-forms#using-the-formbuilder-service-to-generate-controls

Example of using the generated form with ngModel inputs: 
https://angular.dev/guide/forms/reactive-forms#associate-the-formgroup-model-and-view

apps/picsa-apps/dashboard/src/app/modules/crop-information/crop-information.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  public override async init() {
+    await this.supabaseService.ready();
+    await this.listCropProbabilities();
+  }
+
+  public async listCropProbabilities() {
+    const { data, error } = await this.supabaseService.db
+      .table('station_crop_information')
+      .select<'*', ICropInformationRow>('*');
```
chrismclarke, line 35: (+) thanks for making the db lookup type-safe, makes things a lot easier when integrating with the rest of frontend components

apps/picsa-apps/dashboard/src/app/modules/crop-information/crop-information.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (error) {
+      throw error;
+    }
+    this.cropProbabilities = data || [];
+  }
+
+  public async addCropProbability(cropProbability: Partial<ICropInformationRow>): Promise<string> {
```
chrismclarke, line 42: nit(non-blocking) - if you want to be more strict with type definitions the generated types also include separately the data requirements for inserting rows `Database['public']['Tables']['crop_data']['Insert']` - although I don't think a huge deal as we'll likely manage data validation within the form so doubt it would pick up many type errors.

chrismclarke (changes requested):
Thanks again for all the updates. Find a few mostly minor comments in-line, I think the one thing that would be really nice to prioritise would be getting the data input validated using angular forms as they're quite a powerful tool which I'm slowly trying to implement more across the codebase. Otherwise looking good as a basis until we know exactly how we can interact with the climate api to generate probabilities for specific stations (and provide custom crop defintion overrides)

FaithDaka:
@chrismclarke Thank you for all the elaborate feedback in this PR. I've worked on all the requests and fixes.
The pending features include being able to edit and delete an entry but creating, reading are complete. 

https://github.com/e-picsa/picsa-apps/assets/53012890/e6ab643f-4b9e-4d21-a616-c65d17b9ecfc

chrismclarke (approved):
Thanks for carefully addressing all the feedback points, very much appreciated. I think the form validation methods in particular really make the code a lot more maintainable in the long run and also simplifies some of the UI logic.

I still need need to add a global permissions system that provides more fine-grain control on who can/can't edit content, but for now I'll add a quick commit just to make it easier to edit existing rows.

Otherwise I'd say should be good to go!

Minor Updates:
a00447d - automatically navigate back to the crop list page after successful crop add
3ff5559 - additional form validation against db input
283070e - crop entry edit
e75b174 - add common crop picker input

**Updated Demo**
[Screenity video - Mar 11, 2024.webm](https://github.com/e-picsa/picsa-apps/assets/10515065/40bcb4dd-8ba4-48ee-9e7e-b2c01ac8ec8e)

PR #236, opened 2024-02-20
link search route

Pamella014:
## Description

I was able to link the search text to the route navigation where its stored as query params

## Screenshots / Videos

Screenshot from 2024-02-16 09-24-08
[Picsa Extension.webm](https://github.com/e-picsa/picsa-apps/assets/78596562/c8b20ef0-a0cd-4784-aa61-a1c4b31f9ffc)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+yarn lint-staged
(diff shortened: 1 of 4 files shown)


chrismclarke (approved):
Thanks @Pamella014 , the feature works really well! I tested typing into the input to see route param change, reloading page to see search reload success, and clearing input to see param removed.

I've added two small additions
5c5b7c8 - a very common issue when working with rxjs subscriptions is the fact that they should be manually unsubscribed when you no longer need them (can cause small performance issues if too many left open). There's lots of ways to manage this, you can see the code I've added in the linked commit or see more examples at:

54b33d9 - a tiny amount of code tidying for the function that updates the route. It basically uses a small feature/hack that when setting a route param if you set as undefined, e.g. `{searchText: undefined}` the angular router is actually smart enough to just remove that. So I've implemented without need for the conditional statement and renamed the method name to be a little bit more specific

But overall I think the implementation is really nice and uses a few different angular techniques which is good to see you picking up. I'll go ahead and merge here and identify a good next issue to work on.

Oh, and I can see that the video you shared has a link but is not displayed inline. This is an annoying markdown issue, basically you need to leave an empty line before the video to get it to display in github preview. So feel free to update your PR note and add a new line before the video link and it should show correctly

Pamella014:
Thank you Chris.

Pamella014:
I have updated the PR as requested

Issue #219, opened 2024-01-18
bug(android) - farmer activity video playback resume


chrismclarke:
**Is your feature request related to a problem? Please describe.**
When running on web, farmer activity videos play inside a web browser and users can toggle between video and other tabs freely, resuming the video from where they left off whenever the return (example below)

**Web - videos resume correctly**

[PICSA (3).webm](https://github.com/e-picsa/picsa-apps/assets/10515065/c12e576e-127e-4c9a-a387-bab01d9849dc)

However when running on native android device the video opens in a full-screen player, and once exiting loses progress. When trying to replay sometime it will not even load the video, and other times it restarts from the beginning. Ideally the native android version should be able to resume playback in the same way the web player does.

**Android - videos do not resume**

[untitled (1).webm](https://github.com/e-picsa/picsa-apps/assets/10515065/16ae97a4-a34d-47e9-bc19-880932a98421)

**Describe the solution you'd like**
We want to keep full-screen functionality, so ideally somehow hooking into playback events to manually track the time video is paused and automatically resuming from that time in the component

**Additional context**
Code for the video player can be found in `libs\shared\src\features\video-player`. It uses native plugin [capacitor-video-player](https://github.com/jepiqueau/capacitor-video-player)

Running the app on a local android device or emulator for debugging is a bit tricky, however some documentation has been added to https://docs.picsa.app/advanced/testing to try and help organise a good developer workflow (reach out if having issues).

PR #237, opened 2024-02-20
farmer activity video playback resume

OchiengPaul442:
## Description
- This update refines the play method to address a malfunction in the video player. Previously, if a user initiated video playback, terminated the player, and then attempted to replay the video, the player would not operate as expected.
- Focused on rectifying the playback resumption issue for the player on Android devices. This fix ensures a seamless user experience by enabling the video to resume from the point where it was paused, even after the player has been closed.

NB: If the user navigates back to the Farm activity window, the current pause time is reset, and playback starts from the beginning. However, we could potentially use local storage, which is supported on both desktop and mobile platforms, to track the playback times for different videos. This would allow us to resume playback from the last paused position even though the user closed the app.

## Discussion
- This Pull Request is aimed at addressing Issue #219. 

## Screenshots / Videos
https://github.com/e-picsa/picsa-apps/assets/97807374/fb63f335-9ac0-42e1-835b-146f8c45e9c2


chrismclarke (approved):
Thanks for working through all this @OchiengPaul442 , I can appreciate getting an android workflow operational to even recreate the problem was quite a challenge.

The playback resume feature works well, I like your idea of persisting to storage (to potentially track progress across multiple videos and app restarts), I'll write a follow-up issue to avoid this PR getting too cluttered.

I noticed a strange bug when running the videos, perhaps you did too - after playing the video the app gets locked into landscape mode. I've tried adding the [Screen Orientation](https://capacitorjs.com/docs/next/apis/screen-orientation) native plugin to give us a bit more control over the orientation

However.... turns out the video-player plugin had already noted the screen orientation bug and issued a fix (as per `https://github.com/jepiqueau/capacitor-video-player/issues/136`) so I've updated the package, although testing myself I think still need to manually force the orientation to 'unlock' following playback to ensure orientation corrects. Changes pushed in 634fa7c

I also realise now we have multiple developers working on native testing (who will each have their own local IP) it would be really nice to implement a cleaner system for exposing the local development instead of manual comment/edit of the config file. I'll see if I can figure something out a better implementation as a follow-up.

But altogether nice work!

Issue #220, opened 2024-01-18
feat(core) - track custom analytics events


chrismclarke:
**Is your feature request related to a problem? Please describe.**
University team said that they would like to be able to track how many people have been playing the videos and how many people have opened resources.

**Describe the solution you'd like**
Use the existing google analytics integration to track video playback and resource open events as custom analytics events. Ideally this will be done by adding methods to the existing analytics service `libs\shared\src\services\core\analytics.service.ts` to track custom events, and the service called as required from video player and resource service. 

**Additional context**
Documentation for firebase analytics used in the platform can be found at:
https://firebase.google.com/docs/analytics

Code for the video player can be found at `libs\shared\src\features\video-player`

Code for resource open can be found at apps\picsa-tools\resources-tool\src\app\services\resources-tool.service.ts

Issue #241, opened 2024-02-23
feat(video-player): playback service and storage


chrismclarke:
**Is your feature request related to a problem? Please describe.**
Following work on #237 it is now possible for a video to be resumed on native after being closed, however the playback state only persists for as long as the component (so destroyed after page back navigation or app restart).

Instead it would be useful to persist the playback state of all videos across sessions, and provide users indication of which videos have been previously viewed and which are in progress.

**Describe the solution you'd like**
- [x]  Create a `video-player.service.ts` file to manage main logic for storing and retrieving playback state
- [x]  Create a folder in the video-player for `schemas`  and create a database schema to hold video playback information (e.g. video id, current time, total time, playback percentage). Would recommend calling the data table `video-playback`
- [x]  Add data bindings to update db entries from the service when video playback events
- [x]  Add ui to show playback progress bar below start button

**Additional context**
Video player code in `libs\shared\src\features\video-player`

The video player currently generates a random id for each video played, however it would be good instead pass as an input that can be used to link with the db.  E.g. when linking with a video resource the input could allow for the resource.id to be passed
```
<picsa-video-player [source]="videoUri" [id]="videoResource.id"></picsa-video-player>
```

Progress bar ui can just be [mat-progress-bar](https://material.angular.io/components/progress-bar/overview), with some extra css to make it sit below the play start button.

Example of creating db schema and using with service can be found in `apps\picsa-tools\seasonal-calendar-tool\src\app\schema` and `apps\picsa-tools\seasonal-calendar-tool\src\app\services\calendar.data.service.ts`. Full rxdb docs at https://rxdb.info/rx-schema.html

---
**Misc**
I also thought it might be a nice enhancement to look into auto-generating thumbnails for videos, although I think probably better as a lower priority follow-up so excluding from here. But just to put some links for research purposes (so I can close some tabs!):

Web
- https://cwestblog.com/2017/05/03/javascript-snippet-get-video-frame-as-an-image/
- https://immanubhardwaj.medium.com/javascript-how-to-capture-image-from-video-ac7dba01a8e7

Native
- https://github.com/dragermrb/capacitor-plugin-video-editor?tab=readme-ov-file#thumbnailoptions
- https://github.com/sultanmyrza/preview-thumbnail/tree/main

PR #243, opened 2024-02-25
feat: track custom analytics events

OchiengPaul442:
## Description
This PR:
- Implements Google Analytics tracking for video playback functionality in the farm-activity view. This will provide valuable insights into user engagement with the videos.
- Adds Google Analytics tracking for the opening of resources in the resources section. This will help us understand how many users view the resource pdfs on the platform.

## Discussion

This PR resolves issue #220  by implementing the necessary changes.

## Preview
- To preview these changes, please interact with any video in the farm-activity view or open any resource in the resources section and check the corresponding Google Analytics events.

#### Resources Event
image

#### Video Playback Event
image


libs/shared/src/features/video-player/video-player.component.ts
```diff
@@ -74,6 +75,8 @@ export class VideoPlayerComponent implements OnDestroy {
     if (Capacitor.isNativePlatform()) {
       this.initialised = false;
     }
+    // Track video play event
+    this.analyticsService.trackVideoPlay(this.playerId);
```
chrismclarke, line 79: nit(non-blocking): currently the playerId is a randomly generated id, so it doesn't really hold any analytics value. It might be good to implement player IDs as suggested in https://github.com/e-picsa/picsa-apps/issues/241 (but not the full storage solution) so that we can track the id of the resource when passed to the video player and for logging in analytics

This could either be included in this pr (would only be a few lines of code), or as part of a follow-up

apps/picsa-tools/resources-tool/src/app/services/resources-tool.service.ts
```diff
@@ -71,6 +73,8 @@ export class ResourcesToolService extends PicsaAsyncService {
   }
 
   public async openFileResource(uri: string, mimetype: string) {
+    // track the resource open event
+    this.analyticsService.trackResourceOpen();
```
chrismclarke, line 77: nit(blocking): Nice we can track the resource open event but would be even better if we could track the id of the resource being opened. So in order to achieve this we either need to pass an extra argument to the `openFileResource` method (e.g. `id`), or call the tracking function within the `resource-item-file` component itself which does have the id.

I think technically the better solution would probably be passing the id to the service, as that keeps the component more focused on the UI and service on side-effects

libs/shared/src/services/core/analytics.service.ts
```diff
@@ -35,4 +35,20 @@ export class AnalyticsService {
       }
     });
   }
+
+  // Method to track when users play a video
+  public trackVideoPlay(videoId: string) {
+    this.firebaseAnalytics.logEvent({
+      name: 'video_play',
+      params: { video_id: videoId, app_version: APP_VERSION },
```
chrismclarke, line 43: nit(non-blocking): `app_version` will automatically be tracked in various other places so no need to explictily send with each analytics request, but very nice to have the video_id getting passed.

libs/shared/src/services/core/analytics.service.ts
```diff
@@ -35,4 +35,20 @@ export class AnalyticsService {
...
+    });
+  }
+
+  // Method to track when users opens resource file
+  public trackResourceOpen() {
+    this.firebaseAnalytics.logEvent({
+      name: 'open_resource_file',
+      params: { app_version: APP_VERSION },
```
chrismclarke, line 51: nit(non-blocking): same as above, although would also be nice to have the `resource_id` tracked as part of the the payload sent to analytics so that we can track research questions such as which resources are accessed most frequently

chrismclarke (changes requested):
Thanks for the PR @OchiengPaul442 
I can confirm that I'm also able to see the custom analytics events you've started tracking from within the google analytics dashboard so nice that workflow is going as planned.

See a few comments inline, mostly around just trying to increase the amount of info getting tracked in analytics (namely resource and video ids, so that we can have a better insight into the resources users are accessing most frequently)

libs/shared/src/features/video-player/video-player.component.ts
```diff
@@ -1,6 +1,7 @@
 import { Component, ElementRef, HostBinding, Input, OnDestroy } from '@angular/core';
 import { Capacitor } from '@capacitor/core';
 import { ScreenOrientation } from '@capacitor/screen-orientation';
+import { AnalyticsService } from '@picsa/shared/services/core/analytics.service';
```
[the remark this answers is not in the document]
OchiengPaul442: Hey @chrismclarke,  I noticed the build errors earlier and was working on resolving them. Your explanation has provided clarity on the issue, and I’m now in a better position to address it. Thanks

libs/shared/src/services/core/analytics.service.ts
OchiengPaul442: Will work on that.

chrismclarke:
Hey @OchiengPaul442 - I can see you've removed the `work in progress` label which I'm guessing means the PR is ready for second review, although if that is the case would it be possible to be more explicit and use the `re-request review` toggle available in the reviewers pane? Will try take a look tomorrow if that is the case

chrismclarke (approved):
Thanks @OchiengPaul442 , thanks for adding the updates.

I found one issue, there were additional components using the video player that were not passing the resource id, which stopped the video player from working. I fixed the component and added a fallback (console warning + random id) in case other developers use the player without passing an id parameter

I also made a few minor housekeeping improvements, I've prefixed all custom events with `picsa_` to make them easier to locate in the analytics dashboard and moved the analytics enable/disable to the service itself (instead of app component) just to make it more clear for developers

I've included the fix and updates in b55a77f81fab4a589b1e30c9e020e7b6c93d599c

When testing locally I didn't see any analytics events getting tracked, but then I realised that was because I use an adblocker (ublock origin) which automatically blocked GA when running on localhost (so had to disable). I don't think there's anything really to do about this from the code, more just to be aware in case other devs have similar issue.

Oh, and a very minor thing to note, you might see in the udpates I've changed your comment from using single `//` to block `/** */`.
The only reason for this is when using block `/** */` comments vscode intellisense will detect and provide a popup. This works for all variables, functions etc. So I typically use single comments inline in functions when I just want to explain what the code is doing or minor notes, and block comments for names of variables and function descriptions to provide more general info for developers.

See the difference in screenshot below

image

image

But otherwise very nice addition to have, thanks for adding

Issue #248, opened 2024-03-01
feat(dashbboard): ability to upload monitoring form xlsx files and convert to enketo json format


chrismclarke:
**Is your feature request related to a problem? Please describe.**
To display forms in the app there is currently a manual process where a user must first upload to kobotoolbox, deploy to enketo, and then extract the generated json from the enketo website indexeddb entry. This is highly technical and not very administrator friendly.

**Describe the solution you'd like**
- Create a UI component within the dashboard monitoring form page to upload an xlsx file. 
- Deploy cloud function that can convert the xlsx to xlsform
- Use the function to automatically convert when a user uploads an xlsx file and update the database accordingly

The largest part of this is the creation of the cloud function. In order to convert xlsx files to the correct format there are separate scripts that must be executed in python and javascript. An example of the scripts can be found at:
https://github.com/IDEMSInternational/parenting-app-ui/tree/master/packages/@idemsInternational/xlsform-converter

To make matters more challenging, only specific python and node versions are supported, making it hard to use standard cloud function providers (e.g. firebase functions). As such the scripts need to be compiled into a docker image to be deployed to a platform like [Google Cloud Run](https://cloud.google.com/run/docs)

It would be recommended that a standalone git repo be setup to handle the cloud function, based on the example code and then extended to support passing the form in a http request (instead of the example which expects to find the form stored locally).

**Describe alternatives you've considered**
We could try using kobotoolbox rest api to submit form xlsx files and generate enketo links, however this may be difficult to implement and would remove the availability for the dashboard to operate independent of kobotoolbox (which may be desirable for data protection reasons)

**Additional context**
- See existing supabase file upload component at `libs\shared\src\services\core\supabase\components\upload\supabase-upload.component.ts`
- An example xlsx file for conversion can be found in `apps\picsa-server\supabase\data\storage\global\monitoring\forms\widgets.xlsx`

chrismclarke:
Closed by #252

PR #252, opened 2024-03-27
Ft enable xls parsing

khalifan-kfan:
## Description

This PR enables the user to update the Enketo submission form, It mainly solves the issue #248

## Discussion

#248

## Screenshots / Videos

Screenshot 2024-03-27 at 12 14 20
Screenshot 2024-03-27 at 12 14 33


chrismclarke (approved):
Thanks @khalifan-kfan , lots of really great progress here considering all the extra work that's gone into the API endpoints.

I can confirm that the full process is working for me to upload an xls file and have it converted to XML and then enketo form. 

I've added a couple minor additions,
96a7eef - this converts the HTTP request from subscriptions to promise using an rxjs utility method called [firstValueFrom](https://rxjs.dev/api/index/function/firstValueFrom). This is useful in HTTP request where you know you are just going to receive a single response (instead of a stream of data), as you can then use more simple async-await structures. I've also added strong type definitions observerved from the responses

7572c4c - This implements the async-await syntax, and adds an extra method to delete the uploaded file if both steps of the conversion not fully successful (e.g. if an xls file has errors and can't be converted we won't store, so that a user can upload another version)

I also saw then note you made about using storage object ids. A good idea which I had previously tried using, although the challenge is supabase doesn't currently let you set the object id manually so it makes it impossible to import/export data as any time you create a new db all the storage object ids change and any tables referring to them break. So instead I just stored the bucket and folder names as variables to make it easier to retrieve the storage object path and populate as required.

But overall really nice work, there's still a few tasks we'll want to do as a follow-up (e.g. create new form entries, preview the form), I'll plan to create issue for those shortly and happy to merge here for now.

PR #253, opened 2024-03-27
Ft export as image

Pamella014:
## Description

I have added a button to the options screen which will enable the user to export the options as an image

## Feedback Discussion 1

In the issue created you mentioned attaching the image of the crop selected in the download but i cant seem to identify where that happens in the process of creating  an option.

## Feedback Discussion 2

I would also like some clarification on whether the practice is the name of the option that you want extracted as the name of the file when the user downloads the image.

## Screenshots / Videos

Screenshot from 2024-03-27 13-09-36


apps/picsa-tools/option-tool/src/app/pages/home/home.component.scss
```diff
@@ -144,3 +144,7 @@ th.split-values {
   justify-content: center;
   gap: 10px;
 }
+
+.button-margin {
+  margin-left: 1300px;
```
chrismclarke, line 149: nit(blocking) - putting a fixed margin won't work on smaller screens. Normally the solution would be to specify `margin-left: auto` which will use up as much space as possible, however this doesn't work on `mat-button` elements because they're designed to be displayed inline, `display: inline-flex`.

So in order to fix this you should set both `display: block; margin-left:auto` and then to ensure it overwrites the default button styles probably better to put it directly on the html component itself instead of css class, i.e.
```html
<button style="display: block; margin-left: auto"
```
or use a more specific css selector, e.g.
```css
button.button-margin{
    display: block; 
    margin-left: auto
}
```

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
```diff
@@ -27,13 +28,35 @@ export class HomeComponent implements OnDestroy {
   private editorIndex = 0;
   private componentDestroyed$ = new Subject();
 
+  public status = '';
+  public disabled = false;
+  public shareCode: string;
```
chrismclarke, line 33: nit(non-blocking) - I think we are currently only sharing as picture not code so you can probably remove this

apps/picsa-tools/option-tool/src/app/pages/home/home.component.ts
```diff
@@ -27,13 +28,35 @@ export class HomeComponent implements OnDestroy {
   private editorIndex = 0;
   private componentDestroyed$ = new Subject();
 
+  public status = '';
+  public disabled = false;
```
chrismclarke, line 32: nit(blocking) - I see you have used a `disabled` status but it is not linked to the UI. It would be good to rename as `shareDisabled` and link to the share button like
```html
<button mat-button [disabled]="shareDisabled">
```
This will prevent the user repeatedly trying to share while it is preparing. You could also show the status text within the button instead of the word `Share`

chrismclarke (changes requested):
Thanks Pamela, can confirm the core functionality is working well to produce an output image for sharing. See a few minor comments inline, once addressed should be good to go

chrismclarke (changes requested):
Thanks @Pamella014 , can confirm the core functionality is working well to export the image from the options tool for sharing. See minor comments inline, once addressed should be good to go

apps/picsa-tools/option-tool/src/app/pages/home/home.component.scss
Pamella014: thank you for the tip

Pamella014:
I have made the requested changes. Kindly review and let me if everything is working as expected especially showing the status text within the button instead of the word `Share`.

Pamella014:
Screenshot from 2024-03-28 21-43-12

chrismclarke (approved):
Thanks for making all the changes @Pamella014 , can confirm the button working smoothly with disabled state and messages now.

I've pushed a few small additional commits

f299a90 - This uses a (not documented) functionality in the app where any page can dynamically update the main header if they want to include custom content in the header area. I've used this to move your button to the top as a minor style improvement

**Share button now available from main header**
image

51b8b17 - Whenever using the print service it automatically adds a `.print-mode` class to the element in case the developer wants to change any styles for print mode. I've used this to add some minor style improvements, setting some minimum sizes so that it still prints nicely on smaller devices, adds some padding and removes table shadow. See updated example output below

**Print layout includes title, border, padding, page sizing and removes shadow**
Options (2)

Thanks for all the work on this, nice feature to have and for sure will want to continue supporting export and share for all the tools

Pamella014:
Looks great. Thank you @chrismclarke. Will look into the new changes to pick a few lessons.

PR #265, opened 2024-04-22
feat(resources-tool): multiple file download and size display

### LINKED ISSUE #254: feat(resources-tool): download all and size
File resources that appear on a page must be downloaded individually. It would be useful to include a button that can download all

Create a `resource-download-multiple` component that takes an input of resources, displays the total number of resources and total size of resources, and has a button to download all. E.g. (rough mockup, use material styling)

image

If possible it would also be good to update the individual download component to show file size

Image

**Additional context**
See existing code to download single resources at `apps\picsa-tools\resources-tool\src\app\components\resource-download\resource-download.component.ts`
It might be possible to extend the existing code, or might be easier to create a new component

Pamella014:
## Description

I have created a component for downloading multiple files with a single button

closes #254

## Discussion

Feedback 1:  Whether the download button invokes the individual downloads or it downloads all of them at once as seen in the attached video.

Feedback 2: Guidance on how to group the progress icon and the cancel icon because currently they are separated and i cant  seem to find the css.

## Screenshots / Videos
(updated)

[Screenity video - Apr 22, 2024.webm](https://github.com/e-picsa/picsa-apps/assets/10515065/e93cd490-99fc-4fdd-a98d-04ec5dff3987)

---
(original)

[PICSA Manual.webm](https://github.com/e-picsa/picsa-apps/assets/78596562/65e9c2e5-fcd3-4131-94b3-cc64f1f68942)


chrismclarke (commented):
Thanks @Pamella014 

I like the queuing system you've implemented and the general approach.

I've pushed b77d690 to add some minor style improvements on the download buttons (align the sizes below the buttons), and added a shared pipe to handle conversion from KB to MB (rounding as required)

localhost_4200_resources_collection_picsa_videos_picsa_videos_farmer(PICSA-Tablet)

I can see the challenge you mentioned to get the individual resources interacting properly when triggering the download all mechanism - it would be nice if the individual resources showed their own download progress on the page but that's not easily possible given the way each `resource-download` component handles all it's own logic.

I think I might need to refactor that component a little to make the service track the individual downloads, so that when triggered from the multiple component you can see each individual resource spinner go around.

That task will likely have to wait a little, so I think I will plan to tidy up and merge what is here (but probably hide the download all button for now), and then make a follow-up PR or issue once I can get the individual resource component logic sorted out a little

I also noted a separate outstanding issue that the video download buttons don't quite line up because of the way the header is added (can see an outstanding TODO item I left for myself) - will try get that tidied up a bit too

chrismclarke:
But otherwise I'd say mostly there so thank you for the additions, feel free if you want to take on any other issue and I'll just tag you or create a follow-up issue if there's anything here I need you to revisit in the future

PR #290, opened 2024-06-26
Farmer app audio support

FaithDaka:
## Description

- Reusable audio playback service and UI component

## Discussion

- Failure to use local assets

## Screenshots / Videos

Turn your volume on to hear the audio.

https://github.com/e-picsa/picsa-apps/assets/53012890/f6826cb4-8702-4217-892c-494506ae627b


libs/shared/src/features/audio-playback/audio-playback.component.html
```diff
@@ -0,0 +1,8 @@
+<button (click)="togglePlayback()" class="audio-button">
+  <div *ngIf="isAudioPlaying; else playIcon">
```
chrismclarke, line 2: nit(non-blocking) - it would be good for the icon to change when play button first clicked (currently only changes when clicked twice, once to start and once to pause), so that the user gets feedback that the audio click was registered. Could also consider putting the icon in a `mat-button` to get a ripple effect

nit(non-blocking) - feel free if you want to use the `@if(...)` and `@else(...)` syntax. Can also probably just put the if/else around the icon and not need additional ng-template

libs/shared/src/features/audio-playback/audio-playback.component.scss
```diff
@@ -0,0 +1,12 @@
+.audio-button{
+    outline: none;
+    background-color: inherit;
+    border: none;
+    padding: 5px;
+    .mat-icon{
+        color: #7292C7;
```
chrismclarke, line 7: nit(non-blocking): might be good to link the color to the primary/secondary pallette. You could use `var(--color-secondary)` to get the deeper blue, or if you want to register a lighter variant you can define it in `libs\theme\src\themes\_mixins.scss`, testing with something like
```
  --color-secondary-200: #{map-get($accent, 200)};
```
The full list of colors that could be registered can be seen in `libs\theme\src\themes\picsa-default.scss` (currently just the 500 and 50 weight variants are in use)

libs/shared/src/features/audio-playback/audio-playback.component.ts
```diff
@@ -0,0 +1,36 @@
...
+      this.audioService.pauseAudio();
+      this.isAudioPlaying = true;
+    } else {
+      this.audioService.playAudio(this.audioUrl);
+      this.isAudioPlaying = false;
+    }
+  }
+
+  isPlaying(): boolean {
+    return this.audioService.isPlaying();
```
chrismclarke, line 30: nit(non-blocking): I like how you have linked the audio play detection to the service, as it's possible that the service will stop component playback (e.g. if the user tries to start a second audio while the first one is playing), however you might need to remove the `isAudioPlaying` variable from this component to avoid confusion

libs/shared/src/features/audio-playback/audio-playback.service.ts
```diff
@@ -0,0 +1,35 @@
+import { Injectable } from '@angular/core';
+import { Router, NavigationStart } from '@angular/router';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AudioService {
+  private audio: HTMLAudioElement;
```
chrismclarke, line 8: +1 - nice you were able to do get things working with the standard html audio api. In another project I've used [HowlerJS](https://howlerjs.com/), however it looks like the api should be sufficient for now

chrismclarke (changes requested):
Thanks @FaithDaka - all looking good so far.

I tested that the audios could play and also checked adding a second audio element that it would handle stopping one when the next starts (which I think it does as you have created a single audio element within the service, so changing the url kills the previous).

It would be good if you could also add a test div to play audio from a local asset file (like you have added) and check if the implementation works. I've also added a few minor non-blocking comments inline.

FaithDaka:
@chrismclarke The changes you suggested have been made. 

The video shows the fix for the icon bug (it was changing after 2 clicks instead of 1)

https://github.com/user-attachments/assets/481ca24d-bfd5-44ba-acd5-448ce1a50f65

PR #293, opened 2024-07-01
Feat station select

### LINKED ISSUE #276: feat(climate-tool): nearest station select
We are expecting to add a large number of additional stations to the E-PICSA app, and as such selecting the correct station for a user may be difficult. It would be better if the app can automatically recommend which station to use, selecting the station that is closest to the user

When loading the climate too site-select page the app should request the user's approximate location, and use that to calculate which site is closest. This should then automatically pick that station (but still give the user the option to select a different one)

picsa app_climate(PICSA) (1)

**Additional context**
See capacitor documentation for implementing [Geolocation API](https://capacitorjs.com/docs/apis/geolocation)

There are many ways to compute distance between 2 points (e.g. user current location and each station to determine what is closest). The simplest method would be to take the difference in coordinate values to calc longitdue and latitude differences (delta-x and delta-y), and use as 2 sides of a right-angled triangle to calculate the hypotenuse.

image

This won't be entirely accurate due to curvature of the earth, but should be a reasonable estimation. Alternatively could look at exact distances using functions built into leafletJS (e.g. https://leafletjs.com/reference.html#map-distance), or search online for formula (e.g. https://stackoverflow.com/a/51819649). These will be more computationally expensive so probably best to work out simple distances first.

khalifan-kfan:
## Description

This PR adds a feature that calculates the closest distance on the map to a user's current location and highlights is.

## Discussion

Closes #276

## Screenshots / Videos
**Latest**

[Screenity video - Jul 8, 2024 (3).webm](https://github.com/e-picsa/picsa-apps/assets/10515065/39710897-e60f-408c-9d5f-cf9306defd1f)

**Original**
Screenshot 2024-07-01 at 13 28 19


chrismclarke:
Example - using chrome to spoof a different location (for testing)

image

chrismclarke (changes requested):
Just taking a quick look and I can see most of the capacitor 6 packages have been downgraded to capacitor 5 (possibly merging from main didn't resolve correctly). Would you be able to take a look at restoring the package.json as seen in:
https://github.com/e-picsa/picsa-apps/blob/main/package.json

Once restored you can add back the additional package for geolocation, using the v6 version

khalifan-kfan:
@chrismclarke  thanks for this feedback let me look into it

chrismclarke (approved):
Thanks @khalifan-kfan 
I tested the location detection and automatic site lookup and found it working really nicely, good job!

I can see the challenge with trying to render the selected station, been a long time since I looked at the code so struggled a bit also. In the end I remembered that the `picsaMap` viewchild can provide direct access to all the underlying map public functions, so I used this to make a programattic way to update (instead of relying on passed inputs). I think this makes things a little easier, see update in [715920b](https://github.com/e-picsa/picsa-apps/pull/293/commits/715920bb782f2e04c2914b4c63ffb0545e90787a)

I also added a [79233cf](https://github.com/e-picsa/picsa-apps/pull/293/commits/79233cfd5fb5abba1b444d9eae80c7416271a134) to render a custom marker to make it easier to see where the user location is detected as, see example in updated video demo.

Otherwise I think looks good and working well, so should be good to go. Thanks again for all the work on this.

khalifan-kfan:
Yes @chrismclarke  thanks for this [commit](https://github.com/e-picsa/picsa-apps/pull/293/commits/715920bb782f2e04c2914b4c63ffb0545e90787a) it made easer

PR #301, opened 2024-07-16
feat(dashboard): monitoring form preview

Pamella014:
## Description

Form preview on the dash board

## Discussion

Currently the webform isn't displaying on the UI. The issue could be with the way its being imported.

## Screenshots / Videos
**Updated - Example monitoring form displayed in dashboard** 
image

---
Screenshot 2024-07-16 at 10 20 15


chrismclarke (approved):
Thanks again for your work on this @Pamella014 

Following on from the message sent, forms are now rendering which is good, there's still a few ongoing issues related to how some specific HTML content is rendered (e.g. newline characters), but I think probably best that I just try to follow-up with a separate PR to address and merge this for now.

PR #307, opened 2024-07-29
feat: Add photo dialog viewer to photo view component

### LINKED ISSUE #305: feat: photo viewer modal
Following #292 there is now a `photo-list` component that displays a grid of all photos a user has taken. These photos are 120px x 120px to allow for reasonable list display, however are too small to see detail. It would be good if photos could be opened in a larger viewer on tap

Add a dialog/modal template that can display the photo in larger format. This should automatically be opened on tap, and include a button to dismiss (and possibly also move the photo delete button to the modal).

**Alternatives considered**
There are existing packages (e.g. https://www.npmjs.com/package/viewerjs) to handle photo viewing, however these might be a bit overkill for our requirements

**Additional context**
It might be easiest to write the modal template within the same `photo-view` component. You can use [mat dialog](https://material.angular.io/components/dialog/overview) to launch a dialog using local html `ng-template`, see example in `apps\picsa-apps\dashboard\src\app\modules\crop-information\pages\variety-details\variety-details.component.html`

OchiengPaul442:
## Description
- Implemented a mat dialog to display a larger view of selected images.
- Close and delete buttons were added to the dialog for better user interaction.
- Improved mobile responsiveness of the photo view component.

## Issue
- Resolves #305.

## Discussion

## Preview

## Screenshots / Videos
**Latest**

[Screenity video - Jul 31, 2024.webm](https://github.com/user-attachments/assets/b24f4070-fb61-4ebf-a8d6-724605173dde)

---
https://github.com/user-attachments/assets/7f712d87-ce43-40e3-b2bd-a9d644b30ac7

### DIFF
--- a/libs/shared/src/features/photo/components/photo-view/photo-view.component.html
+++ b/libs/shared/src/features/photo/components/photo-view/photo-view.component.html
@@ -1,10 +1,25 @@
 @if(uri()){
-<img [src]="uri()" />
+<img [src]="uri()" (click)="openPhotoDialog()" />
 <!-- TODO - show delete option on long press -->
 <button mat-icon-button (click)="promptDelete()" color="secondary" class="delete-button">
   <mat-icon>close</mat-icon>
 </button>
-
 } @if(errorMsg()){
 <div class="error-msg">{{ errorMsg() }}</div>
 }
+
+<ng-template #dialogTemplate let-data>
+  <div class="photo-dialog">
+    <div
+      mat-icon-button
+      (click)="photoDialog.closeAll()"
+      style="cursor: pointer; width: 100%; display: flex; justify-content: end"
+    >
+      <mat-icon>close</mat-icon>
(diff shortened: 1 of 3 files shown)


chrismclarke (approved):
Thanks @OchiengPaul442 - really nice to have, makes it a lot easier to see the photo details and feels very smooth to use.

I've added 7383b22 to add a few small improvements:
- make the dialog fullscreen (height/width settings can be applied to the dialog container on create)
- remove dialog padding (this isn't as obvious, as the dialog sits outside the template and so needs to use a custom `no-padding` utility class defined in global styles `libs\theme\src\_overrides.scss`)
- remove the delete button from the standard list view (a bit tidier now that we can delete from the popup instead)
- change the input to use signals (as easier to define an input that is required via `input.required`)

See updated video in pr description. But otherwise looking really good, thanks for the addition

Issue #303, opened 2024-07-22
feat(server): automate forecast refresh


chrismclarke:
**Is your feature request related to a problem? Please describe.**
The [Climate API](https://api.epicsa.idems.international/) provides an endpoint to receive latest forecasts for a given country. As the data is updated frequently (multiple times per day) we would like to automate checking on a regular schedule (e.g. hourly) for new data, and populating to the database

**Describe the solution you'd like**
- Create a new edge function that can be triggered to pull forecast data from the API and push to the database
- Create a cron trigger to invoke the function on an hourly basis (might not be possible locally but only once in production, will need to investigate)

**Additional context**
Should note that the specific of the `/forecasts` endpoint is subject to change (as per working document [RFC](https://docs.google.com/document/d/1F-hsU2iXezjVhwGNWEXaPXLAI3_S2NK4RSQbWcchFK8/edit#heading=h.ugf7vsmxta44)), so the focus of initial dev should be more proof-of-concept to get cron scheduled functions working, and then tidied up in the future when the endpoint more stable.

Existing code used clientside at the dashboard to retrieve forecasts and populate to the database can be found in [apps/picsa-apps/dashboard/src/app/modules/climate/climate-api.mapping.ts](https://github.com/e-picsa/picsa-apps/blob/main/apps/picsa-apps/dashboard/src/app/modules/climate/climate-api.mapping.ts#L81).
Essentially we are refactoring this code to run at server-side instead

Working with supabase functions can be a bit tricky as they are written using Deno, so local dev environment will need to install and may want to first look at some basic Deno tutorials. 

Find some additional documentation previously written at: https://docs.picsa.app/server/functions
(will likely need update as we progress to make more developer-friendly)

PR #327, opened 2024-09-20
feat: add forecast updater for climate forecast refresh

OchiengPaul442:
## Description
This PR implements the following changes as part of the proof of concept for automating climate forecast refresh every hour:
### **Created forecastUpdater.ts:**

(created this within the E:\intern\picsa-apps\apps\picsa-server\supabase\functions\climate-refresh for now)
- this edge function pulls climate forecast data dynamically based on the provided country code and stores it in Supabase.
- fetches forecasts from the Climate API and stores the formatted data in Supabase’s climate_forecasts table.
- supports dynamic country_code via URL query parameters.
- uses the upsert method to insert or update forecast data into the database to avoid duplication.
- handles and logs errors during both the forecast fetching and storing processes.

**Hourly  cron trigger**
- needs testing to check the storage of the fetched forecast data before deployment to supabase because it failed on testing this out locally though the function can handle both fetching and storage of the data to supabase.

## Discussion
This PR is solving this issue: #303

### Testing
to test this out locally, have deno installed and run the following command:
`deno run --allow-net --allow-env forecastUpdater.ts`
- when testing locally had to delete the deno.lock file to avoid the hash mismatch errors
- successfully fetched the document/forecast data but could not test the storage locally due to `Error: SUPABASE_URL is required.` 

### **Future Enhancements**
An alternative approach that was found could involve utilizing Supabase's Remote Procedure Call (RPC) to run a stored procedure that fetches and stores the forecast data. This would need setting up an RPC script in Supabase though.

### **some blockers faced**

- When attempting to reuse the existing logic from climate-api.mapping.ts for fetching and storing forecast data, the Deno environment led to several type errors. These errors were due to the lack of alias support in Deno, which would have required significant code refactoring to resolve.

## Preview

## Screenshots / Videos
image


chrismclarke (approved):
Thanks @OchiengPaul442 

Apologies it's taken so long to get back into this project. Looking through the PR I really like how you've managed the different functionalities, particularly running the supabase function server to pass params as required, and mapping the retrieved API data to formats required for the DB.

It's a lot of moving parts so always extra difficult to develop and debug, but at least having everything in one file provides a clear entrypoint. 

Just to comment on a few of the specific issues you've raised

**Sharing code between dashboard scripts and server functions**
You're right, it's particularly difficult to do given the potential for type incompatibilities between deno and browser code targets, particularly how the dashboard code includes lots of imports to services like supabase. Moving forwards we will likely refactor to prioritise calling the supabase functions from the dashboard code, so that we can just have a single source to manage. I'll be working on this in a follow-up PR

**Cron Triggers**
The supabase documentation makes this look fairly straight-forward, but agree will be easier to setup and debug once the main function code operational, so again happy to dive deeper for a follow-up.

**Supabase RPC**
Interesting idea. I think the biggest challenge here is how functions need to be written in SQL or postgres plpgsql, which can be quite complex when compared to typescript. It would be possible to have a hybrid approach where the supabase function just handles API retrieval and postgres function manages db inserts, but honestly I think the JS SDK for supabase is nice enough that keeping it all in the edge functions seems reasonable for me.

Perhaps one place where we might consider more use of postgres functions could be if wanting to automate downloading storage files when new db rows added, but again this could be a follow-up

**General**
I'm not sure if you ran into many other errors when developing, but just to know as part of core updates I've recently updated supabase sdks as well as the core image versions. Also I've started using Deno 2 locally which has some nice benefits such as being able to import npm packages.

But overall this PR is definitely looking like a good start. I'll make a follow-up myself to address some specific app requirements and changes related to the climate API, and create new issues when anything else related to this becomes a priority.

chrismclarke:
Oh, also super interesting to see you have a REST client extension for VSCode. Which one is it? I usually use [Insomnia](https://insomnia.rest) on desktop, but would definitely be interested to try that one too

PR #357, opened 2025-01-30
Feat: show play update prompt

### LINKED ISSUE #348: feat(android): show prompt when app update is available
Many users don't realise an app update is available, and either do not have their devices set to auto-update, or use default auto-update which only happens when connected to wi-fi (not on mobile data). This leaves them stuck on older versions of the app

I think broadly 3 things would be required to provide a good user experience:

1. Show a prompt in app that an update is available when in use
2. Deliver a push notification when wanting users to update in general
3. Show a summary of latest updates/changes in the app when available

For the first piece of this work I think focusing just on (1) would be best (as the others require significant additional work)

**Additional context**
Capacitor plugin that exposes the play app update API
https://github.com/capawesome-team/capacitor-plugins/tree/main/packages/app-update

There is an existing Android plugin that can be used to check the play store for updates and identify if an update is available
[Add any other context or screenshots about the request, or relevant parts of the codebase if known](https://github.com/capawesome-team/capacitor-plugins/tree/main/packages/app-update)

This uses the native android app update API documented at:
https://developer.android.com/guide/playcore/in-app-updates

I expect that using the _flexible_ update strategy would provide the best experience to our users

Example of another repo where this was implemented (likely more complicated than our use requirement):
`https://github.com/IDEMSInternational/open-app-builder/pull/1803`

**Example - notification presented when update available**
image

Pamella014:
## Description

Created an app update service that can be accessed anywhere in the app and then i imported it into the main extension app entry component.

## 
closes #348


apps/picsa-apps/extension-app/src/app/app.component.ts
```diff
@@ -48,6 +50,9 @@ export class AppComponent implements OnInit, AfterViewInit {
     this.resourcesService.ready();
     // eagerly load monitoring service to sync form data
     this.monitoringService.ready();
+    this.ready.set(true);
+    // check for available updates
+    await this.appUpdateService.checkForUpdates();
```
chrismclarke, line 55: nit(non-blocking): even though the check itself is async (so it makes sense to await), you could also call without the await so that the rest of the app can continue to initialise independently of whether we wait for this service or not. 

Sometimes I like to just trigger a function and not wait for the result if happy to let it just happen in the background (no reason to block other functionality)

apps/picsa-tools/extension-content/src/app/pages/home/extension-home.component.ts
```diff
@@ -129,6 +129,7 @@ if (!ENVIRONMENT.production) {
 
 @Component({
   selector: 'extension-home',
+  standalone: true,
```
chrismclarke, line 132: nit(non-blocking): since angular 18 by default all components are standalone, so you no longer need to add this. Can safely remove

libs/shared/src/services/native/app-update.ts
```diff
@@ -0,0 +1,39 @@
...
+      ) {
+        await this.startFlexibleUpdate();
+      }
+    } catch (error) {
+      console.error('Error checking for updates:', error);
+    }
+  }
+
+  /** Starts a flexible update process */
+  private async startFlexibleUpdate() {
```
chrismclarke, line 27: All looks good to me, happy to use the flexible update pathway. I think you can probably remove the console logs as capacitor produces its own logs during AppUpdate operations when running on native device

package.json
```diff
@@ -36,7 +36,6 @@
     "leaflet-draw": "github:enketo/Leaflet.draw#ff730785db7fcccbf2485ffcf4dffe1238a7c617"
   },
   "overrides": {
-    "__note_1__": "See https://github.com/supabase/supabase-js/issues/936",
```
chrismclarke, line 39: nit(blocking): I'm not sure why this got removed but it would be good to keep. It's not actually valid code for the `package.json`, but because you are not able to leave comments this is how I've left a note-to-self to remind why the version of supabase auth is defined the way it is in the block below

package.json
```diff
@@ -73,6 +72,8 @@
     "@capacitor/push-notifications": "^6.0.4",
     "@capacitor/screen-orientation": "^6.0.3",
     "@capacitor/share": "^6.0.3",
+    "@capawesome/capacitor-app-update": "^6.1.0",
+    "@ionic/angular": "^8.4.1",
```
chrismclarke, line 76: nit(blocking): The ionic/angular package is only required if we were using [ionic components](https://ionicframework.com/docs/components). As we are using material design components we don't need these, so the import can be removed (guessing they probably had in a tutorial/example you were reviewing).

chrismclarke (changes requested):
@Pamella014 
Thanks for making the changes to isolate the code in a service, I think overall is looking in a good state. See a few minor comments inline that it would be good to tidy up, and after that I will try to get a testing version deployed so we can confirm if behaves as expected

apps/picsa-tools/extension-content/src/app/pages/home/extension-home.component.ts
Pamella014: I remember getting an error when i was working with the component and adding it seemed to have fixed the error, will look into it again.

package.json
Pamella014: Yes, it was among my first approaches that i made when working on my task. I think i had installed it because i was designed my own dialog notification before realising capacitor has its own inbuilt notifications. I removed the imports in the components but must have forgotten to remove it here.

Pamella014:
Thank @chrismclarke for the feedback. I will review and make the necessary changes. Should i create a new pr once i am done?

Pamella014:
@chrismclarke 
I have made the requested changes and pushed.

chrismclarke (approved):
Thanks for making the updates @Pamella014 , and again apologies for the delayed review.

This will be merged and included in the next release (3.53.0), so it won't be until one more release later that we will be able to see if working as expected or not as it will only show when a new version is available after the user has this version installed. But will keep an eye out for it and let you know if any follow-up required.

Issue #345, opened 2024-12-04
feat(resources): video thumbnail preview


chrismclarke:
**Is your feature request related to a problem? Please describe.**
In both farmer and extension app, whenever a video is shown it uses a blank placeholder for a thumbnail. It would be more informative to show a thumbnail for the video instead. The current player supports the option of providing a custom thumbnail image, however it would be good to auto-generate when not provided

**Describe the solution you'd like**
This may actually be a little more complicated than first thought, as we use

It might first be worth seeing if on web using a standard `<video>` tag might have more support, and updating the component to render differently for web/native
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

For native, we use the plugin https://github.com/harmonwood/capacitor-video-player
This has limited documentation and it's not clear how a thumbnail could be shown, although a workaround might be to show the web player, and when clicked launch the native player instead

**Describe alternatives you've considered**
There are various native android methods to generate video thumbnails, however building a custom plugin would be tricky. E.g.
https://developer.android.com/social-and-messaging/guides/media-thumbnails

If the web implementation doesn't naturally display a thumbnail it might be possible to automatically move the video to a certain time (e.g. 1s) and toggle play/pause to see if that renders an image

**Additional context**
Video player code in `libs\shared\src\features\video-player\video-player.component.ts`

Videos can either be found within the farmer app, or combined in the extension resources, e.g.
https://picsa.app/resources/collection/picsa_videos/picsa_videos_farmer

**Example list of videos without thumbnails**
![picsa app_resources_collection_picsa_videos_picsa_videos_farmer(PICSA-Tablet)](https://github.com/user-attachments/assets/21dffa24-9aea-4248-bce9-65190382ad4b)

**Example - once a video has been played a still image appears. Ideally all should show still images without play**
![picsa app_resources_collection_picsa_videos_picsa_videos_farmer(PICSA-Tablet) (1)](https://github.com/user-attachments/assets/c59e9ea7-36e0-445b-a407-5175cf040398)

chrismclarke:
In case of use, custom plugin written for Android as part of expo (react-native app builder)

https://github.dev/expo/expo/packages/expo-video-thumbnails

https://expo.dev/

chrismclarke:
Closed with #369

PR #369, opened 2025-03-23
Feat: enable video thumbnails

khalifan-kfan:
## Description

This PR adds video thumbnails to downloaded and undownloaded videos

## Discussion

#345

## Screenshots / Videos

Screenshot 2025-03-23 at 14 14 20
Screenshot 2025-03-23 at 14 17 14

Screenshot 2025-03-23 at 14 27 52
Screenshot 2025-03-23 at 17 16 36
Screenshot 2025-03-23 at 17 18 21


apps/picsa-tools/resources-tool/src/app/components/resource-item/file/file.html
```diff
@@ -26,8 +26,13 @@
     </div>
   </div>
 
+  <!-- Add thumbnail status text here, it is only undefined when loading, on failure it would be an empty string  -->
+  @if (thumbnail === undefined && resource.subtype === 'video' && !attachment) {
+  <div  class="thumbnail-status">loading...</div>
+  }
```
chrismclarke, line 32: nit(blocking): I think it would be good to add a component here that can manage the thumbnail, receiving the URL as an input e.g. `<picsa-video-thumbnail [url]="dbDoc._data.url">`

You can also move the logic and styling so that it is easier to adapt and maintain in the future

chrismclarke (changes requested):
Thanks @khalifan-kfan overall I really like this solution to the problem

I particularly appreciate having the flexibility to pick any frame in the video to generate a thumbnail, in the future we could add functionality to the dashboard to let admins specify what time to use and store in the db with the video resource.

I think it would be good if we could try to encaspsulate more of the logic for video thumbnail generation and render in to a shared component (outside of the resources system). You could add a thumbnail component to the shared video player library at `libs\shared\src\features\video-player` , and move the thumbnail generation code to the corresponding service. 

We can also later extend the service to add better URL caching, but for now I think a huge step in the right direction

apps/picsa-tools/resources-tool/src/app/components/resource-item/file/file.html
khalifan-kfan: Alright

khalifan-kfan:
@chrismclarke , thanks for this feedback am going to work on these changes

khalifan-kfan:
Hey @chrismclarke  sorry for the delay but I have committed the changes  as requested

chrismclarke (approved):
Thanks @khalifan-kfan - the code is looking really clean now, thanks for the updates. I've added one minor tidying commit b6cd5bfd3ae6bb0e552bbddc3df8399963615b0f as I noticed the URL could be passed from the resource instead of the dbdoc (so could reduce the number of components the `onlineVideoUrl` had to be passed through)

While testing locally I noted that thumbnails typically took around 1MB of data to generate, just due to the way in which web browsers typically buffer videos that are loading (which happens in the background as you are generating the thumbnail).

This isn't too bad, although I definitely think we will need to raise the priority for storing the thumbnails in the database instead of relying on browser caching (to ensure thumbnails persist across  restarts) - I've added follow-up issue #385 for that

PR #370, opened 2025-03-24
Feat/ track selected page and add a map button 

### LINKED ISSUE #367: feat(climate-tool): Track selected station
Whenever a user opens the climate tool they are prompted to select a station to retrieve data from. The tool will automatically recommend the station closes to their location as determined by GPS, but will need to be re-selected every time they use the tool

1. Update the `ConfigurationService` in `libs\configuration\src\provider.ts` to include saving user settings for the climate tool, namely adding to the `IUserSettings` type
```ts
climate_tool:{
  station_id: string
}
```

2. Use the `ClimateDataService` in `apps\picsa-tools\climate-tool\src\app\services\climate-data.service.ts` to retrieve the climate_tool user settings on load and set the preferred station id. Expose public method to allow the user to set their preferred station id.

3. Update the `SiteSelectPage` to automatically redirect to the saved site if it exists, and if not allow user to select and save site

4. Update the `SiteViewPage` to include the option to re-select from the map, e.g. by including a button with a map-icon next to the station select dropdown

**Additional context**
Example station-select page. This should be skipped if the user has previously set their preferred station
Image

Example site-view page. Should include the option to select from the map as well as dropdown list
Image

Extra care will be needed when managing redirects, as if the map page always redirects to the site-view page then pressing the back button could get the user trapped in a loop where they are always redirected back to site-view. Likely will need to ensure the redirect replaces the current URL instead of adding to nav history.

Similarly, if the site-select page redirects whenever a preferred station has been stored, it will be important to clear the saved user setting when wanting to return to the site-select page so that the user is not automatically redirected

Pamella014:
## Description

The implementation now provides a robust station selection system that remembers user preferred station while allowing easy changes when needed.

## Screenshots / Videos

Screenshot 2025-03-24 at 23 36 38

Closes #367

### DIFF
--- a/apps/picsa-tools/climate-tool/src/app/services/climate-chart.service.ts
+++ b/apps/picsa-tools/climate-tool/src/app/services/climate-chart.service.ts
@@ -58,7 +58,9 @@ export class ClimateChartService {
   ) {}
 
   public async clearChartData() {
-    this.setChart(undefined);
+    this.chartDefinition$.next(undefined);
+    this.chartConfig$.next(undefined);
+    this.chartSeriesData$.next([]);
     this.setStation(undefined);
     this.getPointColour = () => undefined;
   }
@@ -78,8 +80,8 @@ export class ClimateChartService {
     return this.station;
   }
 
-  public async setChart(id?: IChartId) {
-    const definition = id ? this.station?.definitions[id] : undefined;
+  public async setChart(id: IChartId) {
+    const definition = this.station?.definitions[id];
     this.chartDefinition = definition;
     this.chartDefinition$.next(this.chartDefinition);
     if (definition) {
--- a/yarn.lock
+++ b/yarn.lock
@@ -17084,7 +17084,7 @@ __metadata:
 "leaflet-draw@github:enketo/Leaflet.draw#ff730785db7fcccbf2485ffcf4dffe1238a7c617":
   version: 1.0.4
   resolution: "leaflet-draw@https://github.com/enketo/Leaflet.draw.git#commit=ff730785db7fcccbf2485ffcf4dffe1238a7c617"
-  checksum: 10c0/bd20b41e4d1fd552b0c95231ba27f48edec47ea1b32fbf2accbdfaec70845484d15bccd263b23bd1f3af97d9e36588a38b04879d0cdf795678857b1fc7250cca
+  checksum: 10c0/df331497e0ab2708b734ecc8dc5479a7533c32e39f61494d71c2481d6e8b12499d18502c34775d2587a0aeac20366741a3cc06dd9731027af58db9f58040ff28
   languageName: node
   linkType: hard
 


apps/picsa-tools/climate-tool/src/app/pages/site-select/site-select.page.ts
```diff
@@ -57,6 +57,31 @@ export class SiteSelectPage {
     });
   }
 
+  ngOnInit() {
+    this.route.queryParams.subscribe((params) => {
+      if (!params['forceSelect']) {
```
chrismclarke, line 62: Ah, interesting challenge! I can see that you are trying to enable the user to come back to the site select page even when a site has been selected. I think an easier solution would be to just remove the site selected - which you tried to do, but there was a bug so probably why you added this after. 

Not a bad idea at all, but would recommend trying to get the simpler solution to work first before adding a more complicated one

apps/picsa-tools/climate-tool/src/app/pages/site-select/site-select.page.ts
```diff
@@ -66,6 +91,9 @@ export class SiteSelectPage {
   }
 
   goToSite(site: IStationMeta) {
+    this.configurationService.updateUserSettings({
+      climate_tool: { station_id: site.id },
+    });
```
chrismclarke, line 96: nit(non-blocking): I think you took the right approach by adding most of the data manipulation methods to the data service (e.g. set/get preferred station ids), but then you have also replicated the same code in the component - better to just call the service code instead

apps/picsa-tools/climate-tool/src/app/services/climate-data.service.ts
```diff
@@ -28,6 +28,41 @@ export class ClimateDataService {
...
+        console.error(`Saved station ID ${stationId} is invalid.`);
+      }
+    }
+  }
+
+  /** Allow user to set preferred station */
+  public setPreferredStation(stationID: string) {
+    const stationExists = this.stationHashmap()[stationID];
+    if (!stationExists) {
+      console.error(`Station with ID ${stationID} does not exist.`);
```
chrismclarke, line 56: nit(blocking): this is the bug I mentioned before - we don't actually want to check if the station exists as we want a user to just be able to remove their station by setting as something either `undefined` or `null` (or just an empty string). Once users can remove their preferred station we don't need the extra `forceSelect` param

chrismclarke (approved):
Really good attempt to get this feature working. I think there is a little bit of code duplication in some places, but for the most part the user experience is unaffected.

Whilst diving in a bit more deeply I realised one key issue with the approach more generally, and that is we always end up loading the map select page very briefly before reloading, which can cause a bit of screen flicker and console errors as the map is both trying to be created and destroyed at the same time.

A neat way to avoid this is to take the logic you have written and move it to the routing layer so that it can be evaluated before angular tries to load any components. 
Angular has a system of [Route Guards](https://angular.dev/guide/routing/common-router-tasks#preventing-unauthorized-access) which can be used to enable/disable loading of specific routes.

I've updated this PR to use the route guards instead, and tidied up a little to remove some of the duplicate code with [6e452ff](https://github.com/e-picsa/picsa-apps/pull/370/commits/6e452ff43479ed973b04b5e5a864a20d8f967287), and so I think now should be good to go

PR #386, opened 2025-04-25
Feat/individual module  landing page

### LINKED ISSUE #371: feat(farmer-app): individual module landing page
When a user selects a module from within the farmer app they are immediately taken to the first activity within the module (usually an intro video). 

_Home Page_
Image

_Module Home Page_ (starts immediately with content)
Image

Most users will watch the video but not know what to do next (often assuming the module is complete), and so it would be good to provide a better overview of everything in the module before jumping into the first section

A new module home page that appears on every module and provides a quick overview of the content available. Typically this will be along the lines of:
1. Intro Video
2. Interactive Tool
3. Review and summary

Ideally the page should read from the module data to know what is included, and display icon + text of each of the sections

**Additional context**
Current code for the module home can be found in `apps\picsa-tools\farmer-content\src\app\pages\module-home`

Specific module data can be found in `libs\data\farmer_content\data\content`

There might be some additional challenge to manage routing for the videos and tools, as they are currently setup to appear on the module home page. Nested route should be created to display the same content one level deeper.

Pamella014:
## Description

Added icons and text for users to identify available content easily.

## Screenshots / Videos
Screenshot 2025-04-25 at 22 42 24

closes #371


chrismclarke (changes requested):
Thanks for the PR @Pamella014 

The changes to the footer look nice, although I think I possibly didn't communicate well enough in the issue that we actually don't want to be using a footer at all, but instead have buttons on the page that link to each section, more like this:

image

Your PR does however do a good job of highlighting challenges working with the data (particularly how to identify what the main activity/video of each step is). I think the data needs a bit of a refactor first and then quite a bit of work to improve the landing pages.

So for now I'll close this PR and takeover the work as I think it will likely be needed in the next few weeks. Feel free to pick up any of the other open issues instead.

Pamella014:
@chrismclarke thank you for that clarification. Displaying the actual content sounds like a great way to approach it.

PR #399, opened 2025-06-02
feat(video-player): Save video thumbnails to rxdb

### LINKED ISSUE #385: Feat(video-player): Store generated thumbnails in DB
Following work in #369 it's now possible to automatically generate video thumbnail previews of videos before they are downloaded. 

The process relies on streaming in a small amount of the video (usually around 1MB) and using that to export a static frame to use as the thumbnail. 

These thumbnails are stored in-memory, so need to be regenerated any time the browser's in-memory cache is cleared (which can happen quite frequently, particularly on mobile)

Persist generated thumbnails to local indexeddb. The videoplayer service already uses a db schema to store current video playback state to db to allow users to resume videos from where they last reached, so that could be extended to also store thumbnail image data

**Describe alternatives you've considered**
We could consider alternatively generating thumbnails server-side when videos are uploaded and providing the URL so that users only have to download the specific thumbnail image, however integrating with supabase functions would be problematic as it does not support FFMPEG in the deno environment. So we would need to use a 3rd party API or separate google cloud run container to handle generation, which might be a little overkill. 

**Additional context**
Video player db schema available at `libs\shared\src\features\video-player\schema`

Thumbnail component code at `libs\shared\src\features\video-player\video-thumbnail\video-thumbnail.component.ts`

Storing the image data in the db will require converting the binary image to base64 string (and converting back on render). RXDB should have a `blobToBase64String` method that could be used to handle conversion to base64 and then `base64ToBlob` to render in component. 

Alternatively you could try using the `PicsaDatabaseAttachmentService` to store the files within the video schema doc (which should handle file type conversions). `libs\shared\src\services\core\db_v2\db-attachment.service.ts`

khalifan-kfan:
## Description

This PR adds the functionality to save video image thumbnails in rxdb

## Discussion

Closes #385

## Screenshots / Videos

Screenshot 2025-06-02 at 15 58 39


chrismclarke (approved):
Thanks @khalifan-kfan , this is looking really good. 

As discussed I added one small PR just to fix the size of the video and canvas used for generating thumbnails with 9d844b9

I've hardcoded for now as 640x360px which looks pretty good on desktop (could probably go smaller on mobile if we wanted, but feels reasonable). Quickly checking the output stored into indexed and a standard thumbnail is coming in at around 30kb which feels pretty space-efficient 
image

I quickly looked into storing as blobs to disk on native instead of base64, we could probably get around a 30% reduction in size, but would mean when rendering to create and ObjectUrl for the blob and needing to remember to revoke on destroy which gets a little more complicated. So I don't think it's needed right now (there is an existing dbAttachment service which tries to handle in a similar way, but the code could use tidying up and improving for web anyways)

I expect another future use-case will be integrating similar code into the dashboard so that users can manually specify the thumbnail frame when uploading a new video, but also not a priority right now

PR #414, opened 2025-07-06
feat-server remove outdated forecasts monthly

OchiengPaul442:
## Description

- Added a new Supabase Edge Function (forecast-cleanup) to automatically delete daily forecast records and their associated storage files that are older than one month.
- Exposed the function as a new endpoint in the dashboard API.
- Created a database migration to schedule the cleanup function to run automatically on the first day of each month using pg_cron.
- Ensured the cleanup process removes both database entries and storage files to prevent unnecessary storage growth.

## Discussion

## Preview

## Screenshots / Videos
image

image

BEFORE DELTETING

image

AFTER DELETING
image


apps/picsa-server/supabase/functions/dashboard/forecast-cleanup.ts
```diff
@@ -0,0 +1,78 @@
...
+    // Delete from database
+    const { error: dbError } = await supabaseClient.from('forecasts').delete().eq('id', forecast.id);
+
+    if (dbError) {
+      errors.push({ id: forecast.id, error: dbError });
+      console.error(`Error deleting database entry for ${forecast.id}:`, dbError);
+    } else {
+      deletedDbEntries.push(forecast.id);
+    }
+  }
```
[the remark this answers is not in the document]
OchiengPaul442: noted

apps/picsa-server/supabase/functions/dashboard/forecast-cleanup.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  // calculating the date for one month ago
+  const oneMonthAgo = new Date();
+  oneMonthAgo.setMonth(oneMonthAgo.getMonth() - 1);
+
+  // querying the forecasts table for entries older than that date
+  const { data: oldForecasts, error: selectError } = await supabaseClient
+    .from('forecasts')
+    .select('id, country_code, storage_file')
+    .lt('created_at', oneMonthAgo.toISOString());
```
chrismclarke, line 24: optimisation(blocking): As we're not overly concerned about the db storage, I think you can include a filter here to only include entries have have a storage file downloaded (not all db entries have storage files downloaded, particularly when loading older forecasts the db entry is populated and storage file only added on demand).

So you will want some sort of way to ignore cases where the storage_file is `null`. 

In addition, the `created_at` gets automatically populated in the db when received, however it might not actually correctly track the date the forecast was produced (e.g. if I load legacy forecasts from the dashboard the created_at will all have today's date)

So instead I think you will need to query on the ID, which is namespaced to match the google storage bucket where the climate partners store original forecast files, e.g. `20250703/zambia-lunchtime-weather-report-thursday-03-07-2025.html`

So you will probably need to manipulate the isoString, e.g.
`.replace(/-/g,'').substring(0,6)`  to get a prefix like `202506` and then search for ids less than the prefix

chrismclarke (changes requested):
Thanks @OchiengPaul442 - for the most part the code is looking good, although I identified one issue with the query which made not possible to test locally (seeding the db populates forecast `created_at` with today's date so doesn't produce any legacy.

If you're able to make that fix then you should be able to populate legacy forecasts from the dashboard, picking a legacy date from the dropdown and manually clicking the download button to store the forecast file for a couple to test

Image

I was also able to fix the db-seed issue and have updated this branch so hopefully that should work for you to get the dashboard running locally and test whether working as expected

apps/picsa-server/supabase/functions/dashboard/forecast-cleanup.ts
OchiengPaul442: thanks for the clarification @chrismclarke I'll move forward accordingly.

chrismclarke (approved):
Thanks @OchiengPaul442 for the updates and clear demo, I can confirm now also working from my side

Image

The code updates also look good to me so I'd say this one is good to go 🚀 

I'll deploy to prod and let you know if any issues come up

chrismclarke:
Just ran on production, it did have one error which maybe was due to a malformatted legacy file, or perhaps just too large a request when running for the first time...
```
"errors": [
		{
			"id": "db-bulk-delete-failed",
			"error": {
				"message": "<html>\r\n<head><title>414 Request-URI Too Large</title></head>\r\n<body>\r\n<center><h1>414 Request-URI Too Large</h1></center>\r\n<hr><center>cloudflare</center>\r\n</body>\r\n</html>\r\n"
			}
		}
	]
```
But it did clean up all storage files and db entries earlier than June 15 2025, so I think working well. I've enabled the cron schedule so should be good going forwards, thanks again for adding.

I also manually tried running again and all looked good
```
{
	"message": "No old forecasts found for deletion."
}
```

PR #456, opened 2025-08-20
Feat: db core updates

chrismclarke:
## Description

_Summary of main changes_

## Discussion

## Preview

## Screenshots / Videos


chrismclarke:
**Slack**

:warning: DB Breaking Change
I've updated the version of postgres that runs on the server (v15->v17)
To keep local dev environment in sync, I've also updated the local config file that specifies db version when running the supabase server locally.
This will require a full container recreation for any devs running locally, so next time you pull main branch you will need to run the following:
cd apps/picsa-server
npx supabase stop --no-backup
npx supabase start
yarn nx run picsa-server:seed
This should just be a one-time task, moving forwards the regular yarn nx run picsa-server:reset should be sufficient to apply any db changes (just this time that we need to destroy the entire docker container and associated volumes)

PR #612, opened 2026-05-18
fix: functions workspace

chrismclarke:
## Description

- Fix issue where dashboard would fail to list admin users due to cors issue
- Fix functions deployment failure due to changed workspace requirements


chrismclarke:
move into #613

Issue #498, opened 2025-11-25
feat(farmer-app): Offline Resource Sharing


chrismclarke:
**Is your feature request related to a problem? Please describe.**
Internet access is highly limited and variable in some areas, making it difficult to download content and updates required to use the E-PICSA app (notably direct-to-farmer videos). We need some means that an extension officer travelling to an area with limited internet access can share ideally both the app and latest data with anyone that owns a smartphone.

**Describe the solution you'd like**
1. The ability in-app to setup a sharing group that other app users can connect to and share files. This will likely require custom android plugin developed in Java/Kotlin, that uses Android's wifi-connect (and possibly bluetooth fallback) 

This will ideally be created as a separate repo, and published to npm for sharing

 2. In-app mechanisms to search for, connect to and host data sharing groups. The group host should be able to check what files they have downloaded, and provide option to group members to download those files.

3. Possible additional functionality to download entire app bundle for sharing with devices that do not have the app installed (via bluetooth or wifi-direct)

**Describe alternatives you've considered**
Could use only bluetooth, however would be slow for file transfers. Could use only wifi-direct, but may not be supported on some older devices.

Could use qr code generation/scan for setting up and joining groups, but may be confusing if users try to scan qr code using camera instead of dedicated in-app scanner that knows how to process the data

**Additional context**
This is a very large/involved task and so should be broken into smaller tasks when starting development. Would also recommend use of modern ai models to help, particularly given the need to write kotlin code as the fact it will be developed mostly independently to the rest of the app (easier for AI to process)

Android P2P networking docs:
https://developer.android.com/develop/connectivity/wifi/wifi-direct

Capacitor plugin docs:
https://capacitorjs.com/docs/plugins/creating-plugins

Example projects that implements custom plugin in similar way:
- https://github.com/falconeta/capacitor-wifi-connect/blob/main/android/src/main/kotlin/com/falconeta/capacitor_wifi_connect/CapacitorWifiConnect.kt
- https://github.com/saemsplays/wantam/blob/4445998797d3ede70bf6f8fe69f09ba32d9713ce/src/capacitor-plugins/p2p-communication/android/src/main/java/com/ceka/p2pcommunication/P2pCommunicationPlugin.java
- https://github.com/Solez-ai/SwiftDrop/blob/272ffaf90025b832fb2e16493af8323c0885dfb9/android/app/src/main/java/com/swift/drop/WifiP2PPlugin.kt

Example AI initial discussion to plan out feature:
[...]

PR #616, opened 2026-05-30
Feat/offline resource sharing

Pamella014:
## Description
Adds an offline-first share flow for farmer/extension users to share the E-PICSA app and downloaded videos with recipients who may not have internet access.
- New routes under `/farmer/share/*` for installation check, app sharing, Bluetooth sharing, and video sharing
- **Share videos** entry point added to the farmer sidebar
- Share videos list supports multi-select, per-video share, and country-scoped language switching for downloaded locales
- Super app build updated to include farmer-content assets
## Share flow
| Route | Purpose |
|-------|---------|
| `/farmer/share` | Installation check — Yes → videos, No → share app |
| `/farmer/share/app` | Choose Bluetooth or WhatsApp to share the app |
| `/farmer/share/bluetooth` | Share app APK or downloaded videos via Bluetooth |
| `/farmer/share/videos` | Select and share downloaded videos |

## Related Issues

Relates to #498 

### WhatsApp sharing — dedicated screen vs backend?
- Do we need a dedicated WhatsApp screen ?

### Share videos list — downloaded only vs all videos?
- Should the list show **only downloaded videos** (current behaviour), or **all available videos** with undownloaded ones disabled/marked?

## Screenshots / Videos

image

closes #561


chrismclarke:
Likely one more screen `/farmer/share/users` which shows users you can connect to

Once connected, display list of videos for download

UI download progress (later)

Pamella014:
@chrismclarke this is the link to the stitch illustrations we came up with https://stitch.withgoogle.com/projects/17900487470011313545. 

You can also check out this which is where i was trying to come up with a minimalistic version: https://stitch.withgoogle.com/projects/3717412445486592721

So i kind of borrowed ideas from both to come up with the final screens.

